PR #11, opened 2020-06-09
Simplify notebooks

GaelVaroquaux:
Remove a few lines of that are not strictly necessary, partly does to improvements in the underlying libraries

### DIFF
--- a/rendered_notebooks/03_basic_preprocessing_categorical_variables_exercise_02_solution.ipynb
+++ b/rendered_notebooks/03_basic_preprocessing_categorical_variables_exercise_02_solution.ipynb
@@ -79,22 +79,14 @@
    "metadata": {},
    "outputs": [
     {
-     "name": "stderr",
-     "output_type": "stream",
-     "text": [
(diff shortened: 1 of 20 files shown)


rendered_notebooks/03_basic_preprocessing_categorical_variables_exercise_02_solution.ipynb
```diff
@@ -78,23 +78,15 @@
...
     {
      "name": "stdout",
      "output_type": "stream",
      "text": [
       "The different scores obtained are: \n",
-      "[0.87310362 0.87193661 0.87659705]\n",
+      "[0.87398915 0.87194186 0.872543   0.87551188 0.877457  ]\n",
       "The accuracy is: 0.874 +- 0.002\n",
-      "CPU times: user 11.5 s, sys: 249 ms, total: 11.7 s\n",
-      "Wall time: 3.43 s\n"
```
lesteve, line 97: Interesting that some things take twice as much time ... maybe worth a closer look

lesteve:
In an ideal world we would control metadata change in the notebook and the `.py`. For example we would not have the Jupyter kernel name or the full Python version in there because it will likely differ for each person. I have to say I don't know if that's feasible so let's ignore this for now.

GaelVaroquaux:
> Interesting that some things take twice as much time ... maybe worth a closer look

It's not the same hardware: I'm running this on a small laptop with too many firefox tabs open (take an inordinate amount of RAM).

lesteve:
Yeah I completely agree this is a very noisy measurement

Issue #22, opened 2020-06-21
Fix remarkjs-pdf timeout


lesteve:
I was annoyed to have the CI red because of the remarkjs-pdf timeout so I commented it out in https://github.com/INRIA/scikit-learn-mooc/commit/b92c7cf7b1a40fbd16518bd5f30c70b379207532.

If you have time to look at it @brospars maybe by patching remarkjs-pdf to have a bigger timeout (as you did in your PR to remarkjs-pdf) or using decktape to generate the PDF (as mentioned in the remark.js doc [here](https://github.com/gnab/remark#printing)), that would be much appreciated.

_Originally posted by @lesteve in https://github.com/INRIA/scikit-learn-mooc/issues/14#issuecomment-646691960_

brospars:
The issue is deeper than that. Even without timeout overfit.html won't load on chrome (tested locally on chromium). It's due to htmlark inlining all images.

brospars:
If we move all figures to `/figures` and remove `htmlark` it could work.

lesteve:
I am afraid I am not following (probably because this web stuff is not my area of expertise ...), maybe you open a PR and get feed-back from @GaelVaroquaux who did this part originally?

lesteve:
Probably this can be closed since #23 was merged.

PR #23, opened 2020-06-26
Fix slides ci

brospars:
As seen in #22 I had to remove htmlark (images and css inliner) and moves images to `/figures` to prevent chrome webview to time out when exporting to pdf (overfit.html file was to big).


lesteve:
Do I understand correctly that with this PR the slides are not standalone any more?

IIUC @GaelVaroquaux used `htmlark` to have the slides stand-alone ...

https://github.com/lesteve/scikit-learn-tutorial/pull/13

lesteve:
To be honest, I don't really remember why we care about PDF slides so if someone remembers, please be my guest ...

PR #26, opened 2020-07-03
Slide for decision tree

TwsThomas:
This slides shows exemples of : 
- a tree 
- classification
- regression

and also, (not sure if the following shall stand here): 
- Bagging regression / classification
- Boosting regression / classification


glemaitre:
Once that you have a better idea of how long will be the slices, I would not be surprised that we will need to separate trees from ensemble because it might be too long otherwise. I am almost sure that I will have to independent notebook

lesteve:
I pushed this in 8675d54.

Context changing the history auto-closed all the ongoing PRs. It was simpler to merge this manually.

PR #30, opened 2020-07-27
Fix for matplotlib 3.3

lesteve:


lesteve:
This PR:
image

master (from the rendered notebook in the repo assuming it is enough up-to-date)
image

and the CI passes, so merging this one.

PR #35, opened 2020-08-07
First version with Jupyter book

lesteve:
Caveats:
- this is only for the notebooks `0[1-4]*.py` for now. It could easily be extended
- `execute_notebooks` is auto which means that building the jupyter book reruns all the notebooks, which makes it quite painful to work with. I am not sure how to make `execute_notebooks: cache` work with things like using data files e.g. `../datasets/penguins.csv`. It seems like `jupyter-cache` can have assets associated to a notebook (search for `jcache stage add-with-assets` in https://github.com/executablebooks/jupyter-cache), but it does not seem easy to use this from jupyter-book. 

@GaelVaroquaux @ogrisel @glemaitre


lesteve:
OK I am going to merge this one, this first version seems usable with plenty of room for improvement.

GaelVaroquaux:
Wohoo!

Is it easy to expose this on a github pages, using github actions?

lesteve:
There is a github action for this, I can try to set it up. Or do it with CircleCI to be able to look at the HTML in a PR. I'll see.

In the meantime you can have a look at: http://lesteve.github.io/jupyter-book/

PR #39, opened 2020-08-13
Tree - grammar, typos and some re-wording

lucyleeow:
Nice notebook.

I would have liked to see the decision tree plot (with arrows and boxes) earlier and to show it with the plots of decision boundary - it is a nice way to visualise the splits. It also makes it obvious that the splits are done per box/partition (i.e., for the 2nd level you need 2 splits, one for each partition).

I will put specific comments at the relevant sections.

### DIFF
--- a/python_scripts/trees.py
+++ b/python_scripts/trees.py
@@ -187,7 +187,7 @@ print(
 # be able to separate all three species. However, from the previous analysis we
 # saw that by using both features we should be able to get fairly good results.
 # Considering the mechanism of the decision tree illustrated above, we should
-# repeat the partitioning on the resulting rectangle created by the first
+# repeat the partitioning on the resulting rectangles created by the first
 # split. In this regard, we expect that the second partition will be using
 # the feature "Culmen Depth".
 
@@ -369,7 +369,7 @@ print(f"Entropy for partition above the threshold: \n"
 # and Gentoo penguins.
 #
 # With entropy, we are able to assess the quality of each partition. However,
-# the ultimate goal is to evaluate the quality of the over all split and thus
+# the ultimate goal is to evaluate the quality of the overall split and thus
 # combine the measures of entropy in each partition into a single statistic.
 #
 # ### Information gain
--- a/python_scripts/trees.py
+++ b/python_scripts/trees.py
@@ -1,28 +1,28 @@
 # %% [markdown]
 # # Decision tree in depth
 #
-# In this notebook, we will go into details on the internal algorithm used to
-# build the decision tree. First, we will focus on the decision tree used for
-# classification. Then, we will highlight the fundamental difference between
-# decision tree used in classification and in regression. Finally, we will
-# quickly discuss the importance of the hyperparameters to be aware of when
+# In this notebook, we will discuss in detail the internal algorithm used to
+# build the decision tree. First, we will focus on the classification decision
+# tree. Then, we will highlight the fundamental difference between the
+# decision tree used for classification and regression. Finally, we will
+# quickly discuss the importance of the hyper-parameters to be aware of when
 # using decision trees.
 #
 # ## Presentation of the dataset
 #
-# We use the
+# We will use the
 # [Palmer penguins dataset](https://allisonhorst.github.io/palmerpenguins/).
-# This dataset is composed of penguins records and ultimately, we want to
-# identify from which specie a penguin belongs to.
+# This dataset is comprised of penguin records and ultimately, we want to
+# predict the species each penguin belongs to.
 #
-# A penguin is from one of the three following species: Adelie, Gentoo, and
-# Chinstrap. See the illustration below depicting of the three different bird
+# Each penguin is from one of the three following species: Adelie, Gentoo, and
+# Chinstrap. See the illustration below depicting the three different penguin
 # species:
 #
 # ![Image of penguins](https://github.com/allisonhorst/palmerpenguins/raw/master/man/figures/lter_penguins.png)
 #
-# This problem is a classification problem since the target is made of
-# categories. We will limit our input data to a subset of the original features
+# This problem is a classification problem since the target is categorical.
+# We will limit our input data to a subset of the original features
 # to simplify our explanations when presenting the decision tree algorithm.
 # Indeed, we will use feature based on penguins' culmen measurement. You can
 # learn more about the penguins' culmen with illustration below:
@@ -48,7 +48,7 @@ data[target_column] = data[target_column].str.split().str[0]
 data.info()
 
 # %% [markdown]
-# We can observe that they are 2 missing records in this dataset and for the
+# We can observe that there are 2 missing records in this dataset and for the
 # sake of simplicity, we will drop the records corresponding to these 2
 # samples.
 
@@ -57,7 +57,7 @@ data = data.dropna()
 data.info()
 
 # %% [markdown]
-# We will separate the target from the data and we will create a training and a
+# We will separate the target from the data and create a training and a
 # testing set.
 
 # %%
@@ -69,7 +69,7 @@ X_train, X_test, y_train, y_test = train_test_split(
 )
 
 # %% [markdown]
-# Before going into details in the decision tree algorithm, we will quickly
+# Before going into detail about the decision tree algorithm, we will quickly
 # inspect our dataset.
 
 # %%
@@ -81,19 +81,19 @@ _ = sns.pairplot(data=data, hue="Species")
 # We can first check the feature distributions by looking at the diagonal plots
 # of the pairplot. We can build the following intuitions:
 #
-# * The Adelie specie is separable from the Gentoo and Chinstrap species using
+# * The Adelie species is separable from the Gentoo and Chinstrap species using
 #   the culmen length;
-# * The Gentoo specie is separable from the Adelie and Chinstrap species using
+# * The Gentoo species is separable from the Adelie and Chinstrap species using
 #   the culmen depth.
 #
-# ## How decision tree are built?
+# ## How are decision tree built?
 #
 # In a previous notebook, we learnt that a linear classifier will define a
 # linear separation to split classes using a linear combination of the input
 # features. In our 2-dimensional space, it means that a linear classifier will
-# defined some oblique lines that best separate our classes. We define a
-# function below that given a set of data point and a classifier will plot the
-# decision boundaries learnt by the classifier.
+# define some oblique lines that best separate our classes. We define a
+# function below that, given a set of data points and a classifier, will plot
+# the decision boundaries learnt by the classifier.
 
 # %%
 import numpy as np
@@ -146,7 +146,7 @@ plot_decision_function(X_train, y_train, linear_model)
 # %% [markdown]
 # We see that the lines are a combination of the input features since they are
 # not perpendicular a specific axis. In addition, it seems that the linear
-# model would be a good candidate model for such problem, giving a good
+# model would be a good candidate model for such problem as it gives good
 # accuracy.
 
 # %%
@@ -156,10 +156,10 @@ print(
 )
 
 # %% [markdown]
-# Unlike linear model, decision tree will partition the space considering a
-# single feature at a time. Let's illustrate this behaviour by having
-# a decision tree which makes a single split to partition the feature space.
-# the decision tree to make a single split to partition our feature space.
+# Unlike linear models, decision trees will partition the space by considering
+# a single feature at a time. Let's illustrate this behaviour by having
+# a decision tree that only makes a single split to partition the feature
+# space.
 
 # %%
 from sklearn.tree import DecisionTreeClassifier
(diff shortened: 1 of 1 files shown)


python_scripts/trees.py
```diff
@@ -81,19 +81,19 @@
...
-# * The Gentoo specie is separable from the Adelie and Chinstrap species using
+# * The Gentoo species is separable from the Adelie and Chinstrap species using
 #   the culmen depth.
 #
-# ## How decision tree are built?
+# ## How are decision tree built?
 #
 # In a previous notebook, we learnt that a linear classifier will define a
 # linear separation to split classes using a linear combination of the input
 # features. In our 2-dimensional space, it means that a linear classifier will
```
lucyleeow, line 93: This is actually more explanation than is in the linear models notebook. I think I already mentioned it in the other PR but it would be nice explain in linear models that the linear separation line formula is a linear combination of the features.

python_scripts/trees.py
```diff
@@ -81,19 +81,19 @@
...
-# defined some oblique lines that best separate our classes. We define a
-# function below that given a set of data point and a classifier will plot the
-# decision boundaries learnt by the classifier.
+# define some oblique lines that best separate our classes. We define a
```
lucyleeow, line 94: We only demonstrated binary classification in linear model. Maybe we could give more details about how these oblique lines work/look like for >2 classes. e.g., is there one line for each class?

Also maybe explain how these oblique separating lines relate to the decision boundaries..

python_scripts/trees.py
```diff
@@ -156,10 +156,10 @@ def plot_decision_function(X, y, clf, ax=None):
 )
 
 # %% [markdown]
-# Unlike linear model, decision tree will partition the space considering a
-# single feature at a time. Let's illustrate this behaviour by having
-# a decision tree which makes a single split to partition the feature space.
-# the decision tree to make a single split to partition our feature space.
+# Unlike linear models, decision trees will partition the space by considering
+# a single feature at a time. Let's illustrate this behaviour by having
```
lucyleeow, line 160: Very nitpicky but I think we should be consistent in either using British (e.g., 'behaviour', 'modelling', 'organisation') or American ('behavior', 'modeling', 'organization') spelling.
Of course I support the correct and proper way to spell things - British.

python_scripts/trees.py
```diff
@@ -182,13 +183,13 @@ def plot_decision_function(X, y, clf, ax=None):
 )
 
 # %% [markdown]
-# Indeed, it is not a surprise. We earlier saw that a single feature will not
-# help separating the three species. However, from the previous analysis we
-# saw that using both features should be useful to get fairly good results.
+# Indeed, it is not a surprise. We saw earlier that a single feature will not
+# be able to separate all three species. However, from the previous analysis we
+# saw that by using both features we should be able to get fairly good results.
 # Considering the mechanism of the decision tree illustrated above, we should
```
lucyleeow, line 189: 'Considering the mechanism of the decision tree illustrated above' - this is a little confusing to me. mechanism = making splits ? If so this might not be necessary.

python_scripts/trees.py
```diff
@@ -182,13 +183,13 @@ def plot_decision_function(X, y, clf, ax=None):
...
-# repeat the partitioning on each rectangle that was previously created. In
-# this regard, we expect that the partition will be using the feature "Culmen
-# Depth" this time.
+# repeat the partitioning on the resulting rectangle created by the first
+# split. In this regard, we expect that the second partition will be using
```
lucyleeow, line 191: Might be a little confusing, the second partitioning sounds like there will be one 2nd partition. But there will be 2 partitions? One each for the each 'resulting rectangle' ?

python_scripts/trees.py
```diff
@@ -362,24 +363,26 @@ def classification_criterion(labels):
...
-# the different partitions to give us a single statistic qualifying the quality
-# of a split. The information gain is defined as the difference of the entropy
-# before making a split and the sum of the entropies of each partition,
-# normalized by the frequencies of class samples on each partition. The goal is
+# Information gain uses the entropy of
+# the two partitions to give us a single statistic quantifying the quality
+# of a split. The information gain is defined as the difference between the
+# entropy
+# before a split and the sum of the entropies of each partition,
+# normalized by the frequencies of class samples in each partition. The goal is
```
lucyleeow, line 382: It might be nice to give a formula for this? I see it's clear from the code what is happening but I think just a formula (and even an example calculation) means people don't have to read the code.

python_scripts/trees.py
```diff
@@ -320,20 +321,20 @@ def plot_decision_function(X, y, clf, ax=None):
...
+# Chinstrap (while they are almost twice more Gentoo).
 #
-# We should come with a statistical measure which combine the class
-# probabilities together that can be used as a criterion to qualify the purity
+# We should use a statistical measure that uses the all class
+# probabilities, as the criterion to qualify the purity
 # of a partition. We will choose as an example the entropy criterion (also used
 # in scikit-learn) which is one of the possible classification criterion.
 #
 # The entropy is defined as: $H(X) = - \sum_{k=1}^{K} p(X_k) \log p(X_k)$
```
lucyleeow, line 334: I think it would be worth defining what the symbols represent.

python_scripts/trees.py
```diff
@@ -320,20 +321,20 @@ def plot_decision_function(X, y, clf, ax=None):
...
 #
-# For a binary problem, the entropy function for one of the class can be
-# depicted as follows:
+# For a binary problem (e.g., only 2 classes of penguins), the entropy function
+# for one of the class can be depicted as follows:
 #
 # ![title](https://upload.wikimedia.org/wikipedia/commons/2/22/Binary_entropy_plot.svg)
 #
```
lucyleeow, line 340: (Below) Nitpick - I would explain it in terms of one class (as this is what is shown in the plot above) e.g., entropy max when proportion from the class is 50% (as other class is also 50%), min when only class x is present, or only the other class is present (prob=0)

lucyleeow: Also maybe a quick explanation of >2 classes, e.g., (I assume) for 3 classes entropy is max when all 3 classes are 33% ?

python_scripts/trees.py
```diff
@@ -469,16 +473,19 @@ def information_gain(labels_below_threshold, labels_above_threshold):
...
+# slight difference is just due to some low-level implementation details.
 #
 # As we previously explained, the split mechanism will be repeated several
-# times (until we don't have any classification error on the training set). In
-# the above example, it corresponds to increasing the `max_depth` parameter.
+# times (until there is no classification error on the training set,
+# i.e., all final partitions consist of only one class). In
+# the above example, it corresponds to setting the `max_depth` parameter to
+# `None`. This allows the algorithm to keep making splits until the final
+# partitions are pure.
```
lucyleeow, line 483: Tricky but should we be talking about splitting until no error ... (as this would be un-advisable due to overfitting) ? Is this the 'default' action of the tree algorithm?

python_scripts/trees.py
```diff
@@ -626,9 +632,10 @@ def plot_regression_model(X, y, model, extrapolate=False, ax=None):
 _ = plot_regression_model(X_train, y_train, tree)
 
 # %% [markdown]
-# We see that the decision tree model does not have a priori and do not end-up
+# We see that the decision tree model does not have a priori and we do not
```
lucyleeow, line 635: I assume priori = no assumption on data distribution ? 

Maybe make the connection why straight line = no assumption about the data distribution... ?

python_scripts/trees.py
```diff
@@ -626,9 +632,10 @@ def plot_regression_model(X, y, model, extrapolate=False, ax=None):
...
+# end-up
 # with a straight line to regress flipper length and body mass. The prediction
-# of a new sample, which was already present in the training set, will give the
+# of a sample from the training set will give the
```
lucyleeow, line 638: I'm not sure exactly how to interpret the regression tree plot. How can one visualise the splits? Maybe a tree plot alongside would be helpful?

Also it's not obvious to me that predicting using a training sample will give the target of the training sample .. ?

lucyleeow: Below 'In the case of regression, the predicted value corresponds to the mean of the target in the node.' - I don't know how to tell what is a node in the graph.

lucyleeow: Below - 'extrapolate to unseen data' - do you mean extrapolate to unseen data outside of the range seen in the training dataset?

TwsThomas:
> I would have liked to see the decision tree plot (with arrows and boxes) earlier
There is some slides on trees, with that kind of figures, that should be presented before this notebook.

python_scripts/trees.py
TwsThomas: Yes indeed, I think is okay to repeats things.
> it would be nice explain in linear models that the linear separation line formula is a linear combination of the features.

+1

python_scripts/trees.py
TwsThomas: +1

python_scripts/trees.py
TwsThomas: For me also it is not clear what "a priori" means here

TwsThomas: For me also it is not clear what "a priori" means here.

python_scripts/trees.py
TwsThomas: For the moment I'm fine with that. It seems to me it is clearly explain here.

python_scripts/trees.py
TwsThomas: In the code below I can read : 
> and minimum when only samples for a single class is present.

Do you think we should add more details here ?
With 3 classes, it is gonna be very verbose I think.

python_scripts/trees.py
lucyleeow: Sorry i meant that in `linear.py` to include this nice explanation of linear separation.

python_scripts/trees.py
lucyleeow: Opened issue

python_scripts/trees.py
lucyleeow: https://github.com/lucyleeow/scikit-learn-mooc/pull/3

python_scripts/trees.py
lucyleeow: https://github.com/lucyleeow/scikit-learn-mooc/pull/3

python_scripts/trees.py
lucyleeow: v

python_scripts/trees.py
lucyleeow: Made a suggestion in new commit.

python_scripts/trees.py
lucyleeow: It is clearly explained. My concern is that doing this is often not advisable as it leads to overfitting. 

Is it too verbose to say something like:
'until there is no classification error on the training set or one of the limiting parameters such as `max_depth` or `min_samples_leaf` has been reached.'  ?

python_scripts/trees.py
```diff
@@ -340,8 +340,11 @@ def plot_decision_function(X, y, clf, ax=None):
 #
 # Therefore, the entropy will be maximum when the proportion of samples from
 # each class is equal (i.e. $p(X_k)$ is 50%) and minimum when only samples for
-# a single class is present (i.e., $p(X_k)$ is 100%, definitely class `X`,
-# or 0%, definitely the other class).
+# a single class is present (i.e., $p(X_k)$ is 100%, only class `X`,
```
TwsThomas, line 343: I am not sure  to what refer the notation $p(X_k)$.
Is X the data ? I guess 'k' is the class.
Maybe we could use p_k to represent the frequency of the class k in the partition.

lucyleeow: Whoops, I was trying to use the same notation as the graph: https://upload.wikimedia.org/wikipedia/commons/2/22/Binary_entropy_plot.svg
I have amended it to be the same notation as the graph

TwsThomas: We should be careful having the same notation with the entropy formula above

lucyleeow: Happy to make it the same as the formula above. The notation of the graph won't be the same though

lucyleeow: Wait i did use the same notation as above? 
```
The entropy is defined as: $H(X) = - \sum_{k=1}^{K} p(X_k) \log p(X_k)$
```

lucyleeow: Did you change the notation?

TwsThomas: > Did you change the notation?

Yes ^^
I replace p(X_k) by p_k for the probability of class k.

lucyleeow: Sure, I've amended but now it doesn't match notation in the graph

python_scripts/trees.py
```diff
@@ -626,16 +648,16 @@ def plot_regression_model(X, y, model, extrapolate=False, ax=None):
...
-# same target than this training sample. However, having different body masses
+# We see that the decision tree model does not have a priori distribution
+# for the data and we do not end-up
+# with a straight line to regress flipper length and body mass.
+# Having different body masses
 # for a same flipper length, the tree will be predicting the mean of the
 # targets.
 #
 # So in classification setting, we saw that the predicted value was the most
 # probable value in the node of the tree. In the case of regression, the
```
lucyleeow, line 659: Should 'node' be 'leaf' here as well then @TwsThomas ?

PR #42, opened 2020-08-14
Metrics - grammar, typo, some re-wording

lucyleeow:
For me the notation `:math:` does not render properly. I think jupyter notebook uses `$ $` notation?

Nitpick - consistently use either mislabel or misclassify.

We could also more explicitly talk about how accuracy is affected by the threshold picked but ROC-AUC and average recall does not, as calculating involves using many thresholds.

Specific comments in review.

### DIFF
--- a/python_scripts/metrics.py
+++ b/python_scripts/metrics.py
@@ -127,8 +127,8 @@ np.mean(y_test == y_pred)
 
 # %% [markdown]
 # This measure is also known as the accuracy. Here, our classifier is 78%
-# accurate at classifying if subject will give blood. `scikit-learn` provides a
-# function that computes this metric in the module `sklearn.metrics`.
+# accurate at classifying if a subject will give blood. `scikit-learn` provides
+# a function that computes this metric in the module `sklearn.metrics`.
 
 # %%
 from sklearn.metrics import accuracy_score
@@ -136,18 +136,19 @@ from sklearn.metrics import accuracy_score
 accuracy_score(y_test, y_pred)
 
 # %% [markdown]
-# Scikit-learn also has a method built named `score`, built into
-# `LogisticRegression` which computes the accuracy score by default.
+# Scikit-learn also has a method named `score`, built into
+# `LogisticRegression`, which computes the accuracy score.
 
 # %%
 classifier.score(X_test, y_test)
 
 # %% [markdown]
 # ### Confusion matrix and derived metrics
-# The comparison that we did above and the accuracy that we deducted did not
-# take into account the type of error our classifier was making. The accuracy
-# is an aggregate of the error. However, we might be interested in a lower
-# granularity level - to know separately the error for the two following cases:
+# The comparison that we did above and the accuracy that we calculated did not
+# take into account the type of error our classifier was making. Accuracy
+# is an aggregate of the errors made by the classifier. We may be interested
+# in finer granularity - to know independently what the error is for each of
+# the two following cases:
 # - we predicted that a person will give blood but she/he did not;
 # - we predicted that a person will not give blood but she/he did.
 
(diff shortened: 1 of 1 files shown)


python_scripts/metrics.py
```diff
@@ -355,13 +366,13 @@
...
+# This curve is built using the same principle as the precision-recall
 # curve: we vary the probability threshold to compute "hard" prediction and
-# compute the metrics. As with the precision-recall curve as well, we can
+# compute the metrics. As with the precision-recall curve, we can
 # compute the area under the ROC (ROC-AUC) to characterize the performance of
-# our classifier. However, this is important to observer that the lower bound
-# of the ROC-AUC is 0.5. Indeed, we represented the performance of a dummy
-# classifier (i.e. green dashed line) to show that the worse performance
+# our classifier. However, it is important to observer that the lower bound
+# of the ROC-AUC is 0.5. Indeed, we show the performance of a dummy
```
lucyleeow, line 374: Maybe we could expand. I think the ROC-AUC value can be lower than 0.5 but 0.5 represents random chance level. And < 0.5 means the classifier is better at incorrectly classifying samples than chance and you should revert your predictions.

python_scripts/metrics.py
```diff
@@ -586,10 +597,10 @@ def plot_pr_roc_interactive():
 
 # %% [markdown]
 # While some features are numeric, some have been tagged as `category`. These
-# features need to be encoded in a proper way such that our random forest can
+# features need to be encoded such that our random forest can
```
lucyleeow, line 600: Maybe expand that generally all (all sklearn estimators at least) need numerical data and cannot work with string data.

python_scripts/metrics.py
```diff
@@ -586,10 +597,10 @@ def plot_pr_roc_interactive():
...
 # deal with them. The simplest solution is to use an `OrdinalEncoder`.
```
lucyleeow, line 601: Maybe give a quick description of what ordinal encoding is?

python_scripts/metrics.py
```diff
@@ -634,8 +646,9 @@ def plot_pr_roc_interactive():
 
 # %% [markdown]
 # The :math:`R^2` score represents the proportion of variance of the target
-# explained by the independent variables in the model. The best score possible
-# is 1 but there is no lower bound. However, a model which would predict the
+# that is explained by the independent variables in the model. The best score
```
lucyleeow, line 649: The R^2 equation would be nice here

python_scripts/metrics.py
```diff
@@ -634,8 +646,9 @@ def plot_pr_roc_interactive():
...
+# possible
+# is 1 but there is no lower bound. However, a model that predicts the
```
lucyleeow, line 651: We could expand on this. 0 is no variation in target explained (always predict mean target value value) and less than 0 - your model performs worse than just predicting the mean.

python_scripts/metrics.py
```diff
@@ -645,11 +658,11 @@ def plot_pr_roc_interactive():
 dummy_regressor.fit(X_train, y_train).score(X_test, y_test)
 
 # %% [markdown]
-# The :math:`R^2` score gives insights regarding the goodness of fit of the
+# The :math:`R^2` score gives insight into the goodness of fit of the
 # model. However, this score cannot be compared from one dataset to another and
```
lucyleeow, line 662: Can you expand on why? And maybe give an example (e.g., y and log(y) )

python_scripts/metrics.py
```diff
@@ -310,24 +321,24 @@
...
-# curve:
+# The precision and recall metric focuses on the positive class however, one
+# might be interested in the compromise between accurately discriminating the
+# positive class and accurately discriminating the negative classes. The
+# statistics used for this are sensitivity and specificity. Sensitivity is just
+# another name for recall. However, specificity measures the proportion of
+# correctly classified samples in the tnegative class defined as:
+# TN / (TN + FP). Similar to the precision-recall curve, sensitivity and
+# specificity are generally plotted as a curve called the receiver operating
+# characteristic (ROC) curve. Below is such a curve:
```
lucyleeow, line 341: It is in the plot but we could specifically state that we use 1-specificity

python_scripts/metrics.py
lucyleeow: Also we could give more intuition on ROC-AUC:
- When sensitivity = 1-specificity - it means our prediction is equally likely to be a true positive or false positive. Thus it represents chance.
- Ideal curve - as our chance of true positive increases we want our false positives to stay low as possible - we want the line to be as close y axis as possible.

python_scripts/metrics.py
```diff
@@ -6,17 +6,17 @@
 # Machine-learning models rely on optimizing an objective function, by seeking
 # its minimum or maximum. It is important to understand that this objective
 # function is usually decoupled from the evaluation metric that we want to
-# optimize in practice. The objective function serves as a proxy to the
+# optimize in practice. The objective function serves as a proxy for the
 # evaluation metric.
 # FIXME: add information about a loss function depending of the notebooks
 # presented before the notebook about metrics.
 #
```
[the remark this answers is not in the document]
lesteve: I am guessing this is what the FIXME means.

PR #46, opened 2020-08-18
Tree - apply suggestions from PR #39

TwsThomas:
I applied suggestions seen in PR #39


TwsThomas:
Putting this in PR #39

PR #53, opened 2020-09-10
ensemble - some rewording

TwsThomas:


python_scripts/ensemble.py
```diff
@@ -273,7 +277,7 @@ def bootstrap_sample(x, y):
 )
 
 # %% [markdown]
-# Theoretically, 63.2% of the original data points of the original dataset will
+# Empirically, 63.2% of the original data points of the original dataset will
```
lesteve, line 280: I have to say I did not understand this change, maybe @lucyleeow can comment? This is a theoretical thing in the sense that it can be proven right (when n_samples goes to infinity) ? For example: https://stats.stackexchange.com/a/88994

Maybe stay more vague i.e. something like this: "it can be shown that around 63.2% of the ..."

python_scripts/ensemble.py
```diff
@@ -273,7 +277,7 @@ def bootstrap_sample(x, y):
...
 # be present in the bootstrap sample. The other 36.8% are just repeated
```
lesteve, line 281: I unsure about this. 63.2% of the original dataset will be in the bootstrap sample. 36.8% of the original dataset will not be in the bootstrap sample. I don't think it is correct to say that 36.8% are repeated samples but maybe I am missing something ...

lesteve (commented):
Thanks for this, I guess this is coming from https://github.com/lucyleeow/scikit-learn-mooc/pull/1?

python_scripts/ensemble.py
lucyleeow: My original comment was more about how we estimate it. I might be mistaken on this but:

we currently estimate by doing `bootstrap` on a `large_n` of samples

I suggested that doing `large_n` of `bootstrap`'s, and then taking the average of the `large_n` of bootstraps was more intuitive to me.

But this is really a minor point. Either way is fine

glemaitre: If I recall well what I intended to write is that when `n -> inf` then 63.2% of the data in the bootstrap sample are unique which I thought to be a theoretical thing and not an empirical one?

python_scripts/ensemble.py
glemaitre: It might be better to mention that only 63.2% of the samples are unique.

python_scripts/ensemble.py
lucyleeow: > when n -> inf then 63.2% of the data in the bootstrap sample are unique 

This makes things clearer for me, but just my opinion.

GaelVaroquaux, line 280:
```suggestion
# On average, 63.2% of the original data points of the original dataset will
```

GaelVaroquaux (commented):
Merging, this is an improvement.

PR #61, opened 2020-09-17
Reorder text to match code inputs

aboucaud:
The text mentions both the scaling and the creation of the pipeline, which is two different informations in one go and only comes later in the code. 

This PR aims at splitting the text so the flow is improved.

@lesteve @jeremiedbb

### DIFF
--- a/python_scripts/02_basic_preprocessing.py
+++ b/python_scripts/02_basic_preprocessing.py
@@ -206,11 +206,7 @@ print(
 # the predictive performance. Instead of increasing the number of iterations, we
 # can try to help fit the model faster by scaling the data first. A range of
 # preprocessing algorithms in scikit-learn allows us to transform the input data
-# before training a model. We can easily combine these sequential operations
-# with a scikit-learn `Pipeline`, which chain together operations and can be
-# used like any other classifier or regressor. The helper function
-# `make_pipeline` will create a `Pipeline` by giving as arguments the successive
-# transformations to perform followed by the classifier or regressor model.
+# before training a model. 
 #
 # In our case, we will standardize the data and then train a new logistic
 # regression model on that new version of the dataset.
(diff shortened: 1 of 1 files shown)


python_scripts/02_basic_preprocessing.py
```diff
@@ -206,11 +206,7 @@
...
 # preprocessing algorithms in scikit-learn allows us to transform the input data
-# before training a model. We can easily combine these sequential operations
-# with a scikit-learn `Pipeline`, which chain together operations and can be
-# used like any other classifier or regressor. The helper function
-# `make_pipeline` will create a `Pipeline` by giving as arguments the successive
-# transformations to perform followed by the classifier or regressor model.
+# before training a model. 
 #
 # In our case, we will standardize the data and then train a new logistic
 # regression model on that new version of the dataset.
```
jeremiedbb, line 212: This part does not match the code. I'm not sure sure if we should train a log reg on the scaled data before introducing the pipeline or change the formulation

python_scripts/02_basic_preprocessing.py
```diff
@@ -230,6 +226,13 @@
                                  columns=data_train.columns)
 data_train_scaled.describe()
 
+# %% [markdown]
+# We can easily combine these sequential operations
+# with a scikit-learn `Pipeline`, which chain together operations and can be
```
jeremiedbb, line 231:
```suggestion
# with a scikit-learn `Pipeline`, which chains together operations and can be
```

python_scripts/02_basic_preprocessing.py
aboucaud: It still matches the code in the long run. 
It is true that it might sound a bit off but I cannot find a better phrasing right now.

jeremiedbb: yeah me neither. I guess it's fine as is

jeremiedbb (approved):

PR #64, opened 2020-09-29
Add missing jupytext headers.

lesteve:
- fix formatting issue that were causing some cells not to be executed
- use uniform header for all .py files
- rerender notebooks


lesteve:
Note: to update all the headers I used:
```
jupytext --set-kernel python3 python_scripts/*.py
```

This showed there were issues with the formatting in some notebooks (e.g. missing `# %%` that were causing code to be included in markdown cells).

PR #70, opened 2020-10-22
ENH simplify the tree narration removing entropy explanation

glemaitre:
Removing the narration linked to the internal split mechanism.
I added a couple of insights regarding the prediction.

### DIFF
--- a/python_scripts/trees.py
+++ b/python_scripts/trees.py
@@ -198,19 +198,18 @@ _ = plot_tree(tree, ax=ax)
 # %% [markdown]
 # We see that the split was done the first feature `X[0]` (i.e. "Culmen
 # Length"). The original dataset was subdivided into 2 sets depending if the
-# culmen length was inferior or superior to 43.25 cm.
+# culmen length was inferior or superior to 43.25 mm.
 #
 # This partition of the dataset is the one that minimize the class diversities
 # in each sub-partitions. This measure is also known as called **criterion**
-# and it is a parameter that can be set when instantiating the decision tree.
-# In classification, this criterion can either be the Gini impurity or the
-# entropy.
+# and different criterion can be used when instantiating the decision tree.
+# Here, it corresponds to the Gini impurity.
 #
 # If we look closely at the partition, the sample inferior to 43.25 belong
 # mainly to the Adelie class. Looking at the tree structure, we indeed observe
 # 109 Adelie samples. We also count 3 Chinstrap samples and 6 Gentoo samples.
 # We can make similar interpretation for the partition defined by a threshold
-# superior to 43.25 cm. In this case, the most represented class is the Gentoo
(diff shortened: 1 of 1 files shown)


lesteve:
LGTM overall. The things I am wondering about:
- do we want to mention that the criterion can be either Gini or entropy or even skip this (I guess gini is mention in the output of `plot_tree` so maybe you need to mention it somehow)?
- not sure what to do with https://275-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/trees.html#the-regression-criterion. Maybe showing a tree with two splits, 4 splits, etc ... similarly to what is done for classification to build up the intuition that the tree is piece-wise constant and that the split decides where the "jumps" happen. Right now the first plot for regression does not convey that very well:
image
(i.e. the piecewise constantness does not jump at you).
- there is a "information gain" remaining somewhere, probably this can be removed and we can stay vague with a formulation like "best split according to some criterion that can be set"

python_scripts/trees.py
```diff
@@ -185,361 +185,126 @@ def plot_decision_function(X, y, clf, ax=None):
...
+# level of the tree will be using the feature "Culmen Depth".
 
 # %%
-from sklearn.tree import plot_tree
-
-tree = DecisionTreeClassifier(criterion="entropy", max_depth=1)
-tree.fit(single_feature.to_frame(), y_train)
-_ = plot_tree(tree)
-
+tree.set_params(max_depth=2)
```
lesteve, line 283:
```suggestion
tree = DecisionTreeClassifier(max_depth=2)
```

seems less magical this way.

lesteve:
> (I guess gini is mention in the output of plot_tree so maybe you need to mention it somehow)?

Or maybe use `plot_tree(tree, impurity=False)` ...

glemaitre:
> do we want to mention that the criterion can be either Gini or entropy or even skip this (I guess gini is mention in the output of plot_tree so maybe you need to mention it somehow)?

Yes we can simplify. I would just mention that we use the Gini just to point to the tree plot and that's it.

> not sure what to do with https://275-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/trees.html#the-regression-criterion. Maybe showing a tree with two splits, 4 splits, etc ... similarly to what is done for classification to build up the intuition that the tree is piece-wise constant and that the split decides where the "jumps" happen. Right now the first plot for regression does not convey that very well:

Yes I forgot that it was structured like this. Basically, we want to explain how it works in regression. Likewise the classification, we need 2 plots to explain the principle of predicting the mean of each partition which is in line with your comment. I will make this message to the point and reduce a bit the narration.

glemaitre:
OK this should be better now. I apply your suggestion @lesteve

lesteve:
Merging, thanks!

PR #71, opened 2020-10-22
Improve structure of JupyterBook table of contents.

lesteve:
- put intro notebooks into a single chapter "The predictive modeling pipeline".
- put ensemble models in same part as trees


lesteve:
- put intro notebooks into a single chapter "The predictive modeling pipeline".
- put ensemble models in same part as trees

PR #77, opened 2020-10-26
Remove unecessary to_numpy.

lesteve:
Not sure why it was needed at some point, but it seems to be fine now? Opening this PR to check this.


lesteve:
Ooops did not hit enter here, maybe that was to avoid things like this:
image

and then having to explain that scikit-learn accepts pandas dataframes but in most cases discards column information.

glemaitre:
This probably something that we can pass under silence. It takes the Series and it output predictions :)

PR #78, opened 2020-10-26
Add CI status to point directly to CircleCI website.

lesteve:
This would be easier to see the generated website directly in a PR.


lesteve:
Let's merge and see.

PR #87, opened 2020-10-28
Investigate slow notebooks in GH action

### LINKED ISSUE #76: Slow notebooks inside Github Actions
https://inria.github.io/scikit-learn-mooc/notebook_timings.html

https://313-246063957-gh.circle-artifacts.com/0/jupyter-book/notebook_timings.html

| | CircleCI | GH Action |
|-|----------|-----------|
| python_scripts/04_basic_parameters_tuning | 50s | 390s |
| python_scripts/ensemble | 50s | 150s |

maybe @ogrisel has some suggestions on this one (oversubscription due to the number of CPU that is not correctly detected)?

According to: https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources

> Each virtual machine has the same hardware resources available.
> - 2-core CPU
> - 7 GB of RAM memory
> - 14 GB of SSD disk space

lesteve:
Close #76.


lesteve:
I convinced myself that using OMP_NUM_THREADS=1 rather than 2 fixed the problem. I used also OMP_NUM_THREADS=1 in CircleCI for consistency also this does not seem to have much of an effect.

I pushed https://github.com/INRIA/scikit-learn-mooc/commit/8cdff1e9f20f5adc68b2a776951a0748a7b61bb5

PR #95, opened 2020-11-19
FIX use drop=if_binary

glemaitre:
Use `drop="if_binary"` in `OneHotEncoder`.


glemaitre:
OK so we are in trouble with the rare categories. Until we do not fix https://github.com/scikit-learn/scikit-learn/issues/18072, we cannot modify the code.

glemaitre:
Actually, https://github.com/scikit-learn/scikit-learn/pull/16018/files would be a better fix

PR #97, opened 2020-11-19
EXA add exercises for linear models module

glemaitre:


python_scripts/linear_models_sol_01.py
```diff
@@ -0,0 +1,133 @@
...
+# ### Data loading
+
+# %%
+import pandas as pd
+
+data = pd.read_csv("../datasets/penguins.csv")
+# select the features of interest
+feature_names = "Flipper Length (mm)"
+target_name = "Body Mass (g)"
+X = data[[feature_names]].dropna()
```
GaelVaroquaux, line 23: Should we be using another dataset, without na, rather than having to deal with NAs.

glemaitre: Since this is something that we are storing locally, we could always have a version of CSV without the NA.

GaelVaroquaux:
But why not iris?

glemaitre:
My initial reason was that I adapted the tree/ensemble notebook and readapted for the MOOC and thought to only use this dataset. We would need to change the other examples then.

Another thing to keep in mind is that some communities are moving away from `iris`: https://armchairecology.blog/iris-dataset/
I saw it in the gallery of seaborn for instance.

I personally did not look at all the discussion and therefore I do not advocate to change `iris` if others want to keep it that way. However, I was thinking to be proactive in adopting `penguins` instead of `iris` to avoid any changes in the future.

GaelVaroquaux:
The problem is that penguin is not a drop-in replacement of iris, in
particular the NAs get in the way of being didactic.

I'm certainly not wed to iris, but we need something simple enough to be
didactic.

glemaitre:
I moved the issue regarding the dataset to a new issue. We can decide what to do with these datasets. 
Once done I can redo a pass on the notebook. I will continue here to break down the notebook and injecting a couple of exercises.

Issue #98, opened 2020-11-19
Which datasets to use?


glemaitre:
I am opening an issue such that we can settle on the dataset to use within the entire MOOC.
I think that we have a large enough number of notebooks to have a good idea of the type of datasets that are required when presenting the different concepts.
It is a follow-up to https://github.com/INRIA/scikit-learn-mooc/pull/97#issuecomment-730375008

The current state is the following:

#### `adult_census`

* local CSV
* large number of samples
* classification
* numerical + categorical features
* contains NA 
* contains rare categories
* used in all 4 beginners' notebooks (as is)

#### `california_housing`

* sklearn fetcher
* large number of samples
* regression
* all numerical
* no NA
* used in cross-validation (as is), ensemble (as is), linear model (as is) notebooks

#### `penguins`

* local CSV
* small number of samples
* classification
* numerical + categorical features
* contains NA
* used in ensemble (subset of numerical features / drop NA), linear_model (subset of feature / drop NA / regression + classification), trees (subset of feature / drop NA / regression + classification)

#### `blood-transfusion-service-center`

* openml
* classification
* numerical
* no NA
* imbalanced without processing
* used in metrics

#### `Ames`

* openml
* regression
* categorical + numerical
* no NA
* Used only numerical columns -> interest in non-gaussianality of the target

#### synthetic dataset

* numpy
* numerical
* used in feature selection to create large number of synthetic features

#### `make_classification`, `make_moons`, `make_gaussian_quantiles`

* sklearn
* numerical
* used in feature selection to create large number of synthetic features (make_classification)
* used in linear model for 2D intuitive non-llinear datsets.

glemaitre:
I am thinking that it could be worth making a notebook presenting the different datasets.
Similar to a glossary for technical terms, it could be nice I think.

PR #99, opened 2020-11-19
Simplify: numerical-only adult

GaelVaroquaux:
Save a version of adult with only numerical columns

This dataset enables to simplify the first notebook.

It can also be reused in other places, to simplify exercises and examples (cc @glemaitre ) .


python_scripts/02_numerical_pipeline.py
```diff
@@ -28,11 +28,16 @@
 #
 # We will use the same dataset "adult_census" described in the previous notebook.
 # For more details about the dataset see <http://www.openml.org/d/1590>.
+#
+# As Numerical data is the most natural type of data used in machine
```
glemaitre, line 32:
```suggestion
# As numerical data is the most natural type of data used in machine
```

GaelVaroquaux:
The created dataset has no NA. Looking at the list of available datasets ( https://github.com/INRIA/scikit-learn-mooc/issues/98 ) it seems to me that it fills a nice niche.

Unless there is strong outcry, I am planning to merge this tomorrow (not tonight, to give people time to look at decent hours), in the interest of quick iterations.

glemaitre:
> Unless there is strong outcry

Not on my side. My personal preference would have been a middle-ground:

* identify the numerical with dtypes;
* select by column names using pandas.

My reason is that I would expect people to know `pandas` at this stage.

But I am fine with the proposal, it is just lowering the entry bar at the minimum.

GaelVaroquaux:
> My reason is that I would expect people to know pandas at this stage.

Every line of code means a cognitive overload. The more a notebook or an exercise has lines of code, the harder it is for people to focus on the pedagogical point that it is making, in particular beginners.

The principle of reducing lines of code to focus on the message we are trying to pass through is one that we need to apply through the MOOC, unless we are doing an exercise / example for the purpose of teaching to fit everything together.

There is a lot of potential improvement along this line throughout the course.

GaelVaroquaux:
OK, merging this.

PR #105, opened 2020-11-22
More didactic first notebook:

GaelVaroquaux:
- Split out data to simplify the notebook
- Use a nearest-neighbor classifier, which was discussed in the video,
  is actually better than the LogisticRegression (not explained yet),
  and has no warning
- Demonstrate overfit


lesteve:
LGTM

I guess some other notebooks will need to be adapted. A few things I can think of:
- add the explanation about train-test split when introducing it in the next notebook
- change the exercises + correction to use the numeric only dataset
- should we use this numeric only dataset in all of the numeric-only data notebooks or only in this first notebook?
- we probably want to adapt the notebook that talks about scaling numerical features because we were saying something like "we got a warning previously let's see what we can do about it".
- the first notebook on data exploration uses the full dataset, so it is a bit weird to switch back and forth between full dataset, numeric-only dataset (but trust me this is the same as the previous dataset we were looking at I am not lying), and now back to the full dataset but we only select categorical features with pandas. I would argue that using slightly different versions of the same dataset can be confusing, possibly as confusing as using many different datasets.

My personal preference (I do get @GaelVaroquaux point about trying really hard to reduce cognitive load) would be to keep the boiler-plate pandas code to select the columns, maybe removing the `make_column_selector` that we had and doing it by column names instead.

glemaitre:
> should we use this numeric only dataset in all of the numeric-only data notebooks or only in this first notebook?

Actually, most of the in-depth notebook is only using numeric-only datasets

> we probably want to adapt the notebook that talks about scaling numerical features because we were saying something like "we got a warning previously let's see what we can do about it".

I am discussing this issue slightly in the linear model notebook regarding the interaction with the regularization strength as well.

Since we might have a course at the Inria Academy that will use only the 4 first notebook, I think that we need to tackle this issue within this introductory notebook. I am always concerned that we get some course where the following is happening: https://github.com/scikit-learn/scikit-learn/issues/18824

For sure, the viable solution is to be implemented in scikit-learn avoiding these issues. But in the meanwhile, I would be in favour to have it somewhere in this first part.

GaelVaroquaux:
>     we probably want to adapt the notebook that talks about scaling numerical
>     features because we were saying something like "we got a warning previously
>     let's see what we can do about it".

> I am discussing this issue slightly in the linear model notebook regarding the
> interaction with the regularization strength as well.

OK, I'll let you adapt.

> For sure, the viable solution is to be implemented in scikit-learn avoiding these issues. But in the meanwhile, I would be in favour to have it somewhere in this first part.

Right, but later. Not in the first notebook that runs machine learning.

lesteve:
>> should we use this numeric only dataset in all of the numeric-only data notebooks or only in this first notebook?

> Actually, most of the in-depth notebook is only using numeric-only datasets

Sorry I was not explicit enough, I was talking here about adult-census-numeric-all.csv (numeric-only dataset) vs adult-census.csv (full dataset). And by numeric-only notebooks I was talking about the `02_numerical_pipeline_*` notebooks.

lesteve:
And to be honest, thinking a bit more about it, I am now wondering whether rather than changing the existing Predictive Modeling Pipeline notebooks that have some internal consistency, it would not be easier to add a separate simple notebook maybe just after the ML concepts slides.

An an exercise, I am thinking we could do the `n_neighbors=1` 100% training accuracy thing that is mentioned quickly in the slides.

GaelVaroquaux:
> And to be honest, thinking a bit more about it, I am now wondering whether rather than changing the existing Predictive Modeling Pipeline notebooks that have some internal consistency, it would not be easier to add a separate simple notebook maybe just after the ML concepts slides.

I decided that we needed to move forward. We can always come back later.
These notebook need rework anyhow, because they need much more text (in a MOOC, we are not there to make comments, all the comments must be written).

But also, I found that it was easier to adapt these notebooks 

> An an exercise, I am thinking we could do the n_neighbors=1 100% training accuracy thing that is mentioned quickly in the slides.

+1 for such an exercise!

PR #106, opened 2020-11-24
ENH split and add example in feature selection

glemaitre:


lesteve:
You are on a roll :muscle: !

Maybe as a general small comment we should use different markers than TODO to indicate where the person should write his solution. It makes it hard to grep for TODO when looking for TODO that I may have left in the text to fix them later. Maybe a "Write your code here" or something better. Or I could start using FIXME rather than TODO as well ...

Also maybe keep some exercises for evaluation i.e. without solution :wink:

glemaitre:
> Maybe as a general small comment we should use different markers than TODO to indicate where the person should write his solution. It makes it hard to grep for TODO when looking for TODO that I may have left in the text to fix them later. Maybe a "Write your code here" or something better. Or I could start using FIXME rather than TODO as well ...

Race condition. Yep, I will merge this PR and make an overall change (I was using FIXME within notebooks :)). But this is friendly to have a "Write your selection here" :)

glemaitre:
> Also maybe keep some exercises for evaluation i.e. without solution

Yep. Sometimes, instead of presenting a concept, I took the choice to make it as an exercise indeed. It requires to not be too complicated but it goes away from "copy the notebook snippet course code and change the estimator" that people complain about sometimes. Now they will complain that it is too difficult :)

PR #114, opened 2020-11-27
MNT split and add exercise in cross-valdiation

glemaitre:


GaelVaroquaux:
Should this stay as a Draft PR, or is it sufficiently advanced that we merge it, even if it is imperfect?

GaelVaroquaux:
Given that we are currently working at a rapid pace, I am in general in favor of early merging when possible.

python_scripts/cross_validation_grouping.py
```diff
@@ -0,0 +1,169 @@
+# %% [markdown]
```
GaelVaroquaux, line 1: Should we have a title? It seems so to me.

python_scripts/cross_validation_ex_05.py
```diff
@@ -0,0 +1,51 @@
+
```
GaelVaroquaux, line 1: I think that we need a bit of text in the beginning of the file to describe what this is about

glemaitre: Yes I am working on the example right now. more to come ;)

PR #123, opened 2020-12-01
ENH specify categories and use drop=if_binary

### LINKED ISSUE #120: Specify categories rather than handle_unknown for OneHotEncoder?
This would make it easier to switch from OneHotEncoder to OrdinalEncoder. At the moment probably because of historical reasons we have a mix of `handle_unknown` and `categories`:

```
❯ git grep -iP 'onehotencoder.+categories' python_scripts/
python_scripts/04_parameter_tuning_sol_02.py:categorical_processor = OneHotEncoder(categories=categories)
```

```
❯ git grep -iP 'onehotencoder.+handle_unk' python_scripts/
python_scripts/03_categorical_pipeline.py:    OneHotEncoder(handle_unknown='ignore'),
python_scripts/03_categorical_pipeline_column_transformer.py:    ('one-hot-encoder', OneHotEncoder(handle_unknown='ignore'),
python_scripts/03_categorical_pipeline_ex_02.py:# `OneHotEncoder(handle_unknown="ignore", sparse=False)` to force the use a
python_scripts/03_categorical_pipeline_sol_01.py:    OneHotEncoder(handle_unknown="ignore"),
python_scripts/03_categorical_pipeline_sol_02.py:# `OneHotEncoder(handle_unknown="ignore", sparse=False)` to force the use a
python_scripts/03_categorical_pipeline_sol_02.py:     OneHotEncoder(handle_unknown="ignore", sparse=False),
```

glemaitre:
close #120 

Specifying the columns in both `OrdinalEncoder` and `OneHotEncoder` for consistency.
Since we do so, we can use `drop="if_binary"` in `OneHotEncoder`.

I also make some small improvement regarding some description.

In the future:

* `OrdinalEncoder` will handle rare categories.
* `OneHotEncoding` should handle `handle_unknown="ignore"` as well as `drop="if_binary"`.

### DIFF
--- a/python_scripts/03_categorical_pipeline.py
+++ b/python_scripts/03_categorical_pipeline.py
@@ -217,7 +217,8 @@ from sklearn.pipeline import make_pipeline
 from sklearn.linear_model import LogisticRegression
 
 model = make_pipeline(
-    OneHotEncoder(categories=categories), LogisticRegression())
+    OneHotEncoder(categories=categories, drop="if_binary"),
+    LogisticRegression())
 
 # %% [markdown]
 # Finally, we can check the model's performance only using the categorical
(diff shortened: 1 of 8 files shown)


lesteve:
Nice I did not think of the combination `drop='if_binary'` + `categories=categories`. This makes it a lot simpler indeed.

PR #141, opened 2020-12-11
MNT move the parameter tuning in a separate module

glemaitre:
Move the section about hyperparameter tuning as a single module.


jupyter-book/_toc.yml
```diff
@@ -56,6 +47,18 @@
         - file: python_scripts/cross_validation_sol_01
     - file: overfit/overfit_take_away
 
+- part: Fine tune your model
+  chapters:
+    - file: tuning/parameter_tuning_module_intro
+    - file: python_scripts/04_parameter_tuning
```
ogrisel, line 53: We could rename this section `tuning/manual_parameter_tweaking` to contrast it with `tuning/automated_parameter_tuning` for the next notebook (and change their titles accordingly). WDYT?

ogrisel: we can remove the `04_` prefix now.

glemaitre: I would prefer to use `parameter_tuning_manual` to use `parameter_tuning_` as a prefix if possible

PR #155, opened 2021-01-15
Pin nbformat to 5.0.8 to avoid regression.

lesteve:
nbformat 5.1 adds random ids to cells which creates issues with jupyter-cache.

This manifests itself as cell outputs not being shown (e.g. `df.head()` output is blank below):
image

Warnings like:
```
WARNING: Notebook code has no file extension metadata, defaulting to `.txt`
WARNING: Couldn't find cache key for notebook file python_scripts/02_numerical_pipeline_ex_01.py. Outputs will not be inserted.
```

Thanks @Phaide for noticing!

This problem is mentioned here for example: https://github.com/mwouts/jupytext/issues/715

### DIFF
--- a/.github/workflows/deploy-gh-pages.yml
+++ b/.github/workflows/deploy-gh-pages.yml
@@ -26,7 +26,9 @@ jobs:
       run: |
         pip install -r requirements.txt
         pip install jupyter-book
-
+        # nbformat 5.1 adds random id which creates problems with jupyter-cache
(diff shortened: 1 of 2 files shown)


lesteve:
Outputs are shown: https://762-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/02_numerical_pipeline_introduction.html

Merging!

PR #157, opened 2021-01-16
Multiple rewordings

Phaide:


lesteve:
Thanks for this!

Reminder about the quizz workflow don't change quizzes directly in this repo.

A bit cumbersome but the way we are doing it:
- update quizzes in https://gitlab.inria.fr/learninglab/mooc-scikit-learn/mooc-scikit-learn-coordination/-/tree/master/jupyter-book (commit and push).
- Then use the `convert_exercises.py` script to update quizzes in the github repo from those in the gitlab repo github https://github.com/INRIA/scikit-learn-mooc/blob/master/jupyter-book/convert_exercises.py

lesteve:
@ogrisel has started working on reviewing the quizzes in the gitlab repo, maybe try to coordinate with him for the changes in the quizzes.

Here is the diff restricted to the quizzes:

`````diff
diff --git a/jupyter-book/linear_models/linear_models_quiz_02.md b/jupyter-book/linear_models/linear_models_quiz_02.md
index 0a7c729..3557cc3 100644
--- a/jupyter-book/linear_models/linear_models_quiz_02.md
+++ b/jupyter-book/linear_models/linear_models_quiz_02.md
@@ -1,7 +1,7 @@
 # ✅ Quiz
 
 ```{admonition} Question
-Is it possible to fit a linear model on a dataset where the data have a
+Is it possible to fit a linear model on a dataset where the data has a
 non-linear link with the target?
 
 - a) yes
diff --git a/jupyter-book/linear_models/linear_models_quiz_03.md b/jupyter-book/linear_models/linear_models_quiz_03.md
index c80d723..5c030f7 100644
--- a/jupyter-book/linear_models/linear_models_quiz_03.md
+++ b/jupyter-book/linear_models/linear_models_quiz_03.md
@@ -5,7 +5,7 @@ Regularization refers to:
 
 - a) constraining the intercept of the linear model to be zero
 - b) shrinking the weights of the linear model towards zero
-- c) use a subset of the available data when fitting a linear model
+- c) using a subset of the available data when fitting a linear model
 ```
 
 +++
diff --git a/jupyter-book/linear_models/linear_models_quiz_04.md b/jupyter-book/linear_models/linear_models_quiz_04.md
index 4c14df3..fde8a8c 100644
--- a/jupyter-book/linear_models/linear_models_quiz_04.md
+++ b/jupyter-book/linear_models/linear_models_quiz_04.md
@@ -1,7 +1,7 @@
 # ✅ Quiz
 
 ```{admonition} Question
-By default, a `LogisticRegression` in scikit-learn apply:
+By default, a `LogisticRegression` in scikit-learn applies:
 
 - a) no penalty
 - b) a l2 penalty
diff --git a/jupyter-book/overfit/learning_validation_curves_quiz.md b/jupyter-book/overfit/learning_validation_curves_quiz.md
index 2351733..07d7334 100644
--- a/jupyter-book/overfit/learning_validation_curves_quiz.md
+++ b/jupyter-book/overfit/learning_validation_curves_quiz.md
@@ -27,8 +27,8 @@ For a fixed training set, if we change a model parameter to give the model more
 flexibility, are we more likely to observe:
 
 - a) a wider difference between train and test errors
-- b) a reduction in the difference between train and test errors
-- c) a increase in the train error
+- b) a reduction of the difference between train and test errors
+- c) an increase in the train error
 - d) a decrease in the train error
 ```
 
@@ -39,8 +39,8 @@ For a fixed choice of model parameters, if we increase the number of labeled
 observations in the training set, are we more likely to observe:
 
 - a) a wider difference between train and test errors
-- b) a reduction in the difference between train and test errors
-- c) a increase in the train error
+- b) a reduction of the difference between train and test errors
+- c) an increase in the train error
 - d) a decrease in the train error
 ```
 
diff --git a/jupyter-book/predictive_modeling_pipeline/02_numerical_pipeline_quiz.md b/jupyter-book/predictive_modeling_pipeline/02_numerical_pipeline_quiz.md
index 813d361..3679916 100644
--- a/jupyter-book/predictive_modeling_pipeline/02_numerical_pipeline_quiz.md
+++ b/jupyter-book/predictive_modeling_pipeline/02_numerical_pipeline_quiz.md
@@ -64,7 +64,7 @@ Cross-validation allows us to:
 +++
 
 ```{admonition} Question
-`make_pipeline` (as well as `Pipeline`) is used for:
+`Pipeline`s are used to:
 
 - a) run cross-validation
 - b) combine one or several transformers and a predictor
diff --git a/jupyter-book/predictive_modeling_pipeline/03_categorical_pipeline_quiz.md b/jupyter-book/predictive_modeling_pipeline/03_categorical_pipeline_quiz
index 498a87a..063b62a 100644
--- a/jupyter-book/predictive_modeling_pipeline/03_categorical_pipeline_quiz.md
+++ b/jupyter-book/predictive_modeling_pipeline/03_categorical_pipeline_quiz.md
@@ -15,7 +15,7 @@ Ordinal categories correspond to:
 
 - a) categorical variable with few categories
 - b) categorical variable with meaningful order
-- c) categorical variable with only number
+- c) categorical variable with only numbers
 ```
``````

File: jupyter-book/linear_models/linear_models_module_intro.md [hunk 1]
```diff
@@ -4,10 +4,14 @@
 
 <!-- Give in plain English what the module is about -->
 
-This module will present more in details models using a linear parametrization.
+In this module, will go further into details regarding models that use 
+linear parametrization.
+\
```
lesteve, line 9: Hmmm those are real `\` ... can I ask why, maybe a quirk of your editor, maybe a subtle markdown thing? 

Unless there is a good reason I am missing, I think they should be removed (here and in the other files)

jupyter-book/linear_models/linear_models_quiz_02.md
```diff
@@ -1,7 +1,7 @@
 # ✅ Quiz
 
 ```{admonition} Question
-Is it possible to fit a linear model on a dataset where the data have a
+Is it possible to fit a linear model on a dataset where the data has a
```
lesteve, line 4: So some people insist `data` is plural (datum is the singular). Honestly I couldn't care less about this, I think we should leave it as is, there are higher priorities.

And sorry since I couldn't care less I don't even think we try to make this consistent in the MOOC

jupyter-book/overfit/overfit_take_away.md
```diff
@@ -6,23 +6,27 @@
   **noise** in the data, and the **high flexibility** of common machine learning
   models.
 
-- **Underfitting** happens when the learned prediction functions make
-  **systematic errors**. This can be caused by a choice of the model family and
-  parameters that leads to a **lack of flexibility** to capture the repeatable
+- **Underfitting** happens when the learnt prediction functions suffer from
```
lesteve, line 9: learned vs learnt: https://www.grammarly.com/blog/learned-learnt/

learned didn't shock me at all, I think we are learning more towards US English in this course.

Same thing as above though, let's keep it as it is, higher priorities ...

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -93,41 +96,55 @@
...
+# The next columns contain related categorical data.
+# These are discrete ; in contrast with continuous, they can only take a
+# value from a finite set, which we call categories.
+# We will come back later on this type of data and how to handle them.
+# Here, we are only interested in recognizing them.
 #
-# Here, the column `"education-num"` gives an example. The number encode the
-# education level which can only correspond to specific values. We can quickly
-# check the number of occurrence of each category to get convinced.
+# Let’s take for example “education-num”.
```
lesteve, line 120: Let's avoid quirky `"` (i.e. different one for opening and closing) character ...

python_scripts/03_categorical_pipeline.py
```diff
@@ -52,38 +53,37 @@
...
-# useful tip is to look at the data type of each column.
+# Now the question is: how can we easily recognize categorical columns
+# among the dataset ? Part of the answer lies in the columns' data type:
 
 # %%
 data.dtypes
 
 # %% [markdown]
-# If we look at the "native-country" column data type, we observe it is of
-# `object` data type. The reason is that categories are represented with
```
lesteve, line 63: Hmmm maybe keep this, some people will not be very familiar with numpy. Adding string may be make it more clear than only dtype object

python_scripts/03_categorical_pipeline.py
```diff
@@ -93,9 +93,10 @@
 categorical_columns
 
 # %% [markdown]
-# We can create the selector by passing the data type to include or exclude.
-# Once the selector created, we can pass the input dataset and the name of
-# the selected columns will be returned.
+# Here, we created the selector by passing the dtype to include ;
```
lesteve, line 96: Honestly I am not a super big fan of `;` we are not writing a novel ...

Only my opinion but shorter sentences and `.` make it easier to follow.

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -93,41 +96,55 @@
...
+# level. Therefore, there is only a specific amount of values each sample can
+# take. Let’s print out the distribution to get an intuition of the data:
 
 # %%
-data["education-num"].value_counts().sort_index()
+distribution = data["education-num"].value_counts().sort_index()
+print(distribution)
+
+# %%
+import matplotlib.pyplot as plt
```
lesteve, line 130: Curious, what makes does the plot bring?

IIRC the point here is to show that there is a finite number of possible values, not really look at the distribution.

Also we have kind of avoided in the first notebooks to use detailed matplotlib code (I think not 100% sure). I guess using pandas to do a plot in one-line could be better if we really want to keep the plot.

python_scripts/03_categorical_pipeline_column_transformer.py
```diff
@@ -82,10 +82,10 @@
 #   information will be correlated;
 # * **numerical scaling** numerical features which will be standardized.
 #
-# We can now create our `ColumnTransfomer` by specifying a list of triplet
-# (preprocessor name, transformer, columns). First, let's start by creating
-# a transformer for the numerical and categorical part as we did in the
-# previous notebooks.
+# We can now create our `ColumnTransfomer` by specifying three values:
```
lesteve, line 85: Yeah this is a bit weird here we are not creating the `ColumnTransformer` yet maybe we should say that we are creating the preprocessors for each type of column.

Then in the next cell, we say that the `ColumnTransformer` takes a list of (preprocessor_name, transformer, columns) or something like this.

lesteve (commented):
First pass.

python_scripts/03_categorical_pipeline.py
lesteve: Ignore this comment for now, I needed to rant a little bit ...

jupyter-book/linear_models/linear_models_module_intro.md
Phaide: Hi

It's a subtle markdown thing ;)

It allows to have line breaks without space between two lines, and is used to break down massive paragraphs.

For instance:

This paragraph uses the anti-slash trick,\
and while there is a line break, there is no extra-space in-between.

```
This paragraph uses the anti-slash trick,\
and while there is a line break, there is no extra-space in-between.
```

Another way of doing it: using double spaces at the end of the line, but it is subtle.  
See https://gist.github.com/shaunlebron/746476e6e7a4d698b373

While it is not part of the Markdown specifications, it is included in most markdown readers.

python_scripts/02_numerical_pipeline_hands_on.py
Phaide: Nice catch, a quirk of my editor...

jupyter-book/linear_models/linear_models_module_intro.md
lesteve: Interesting I did not know this. I would favor a blank line, this seems so much simpler ...

python_scripts/02_numerical_pipeline_hands_on.py
Phaide: Right, I didn't see it that way.

I think I wanted to emphasize on the unequal distribution.

Next paragraph:
```
# We can observe that the distribution is very unequal between categories.
# Such columns would require a specific processing which is different from
# continuous values, so we will ignore them for now.
```

jupyter-book/linear_models/linear_models_module_intro.md
lesteve: <rant>Also I am not a big fan of back-slashes in general</rant> :smirk:

lesteve: (maybe because I had to work on Windows at one point, I don't know ...)

jupyter-book/overfit/overfit_take_away.md
Phaide: I agree, and while both are valid, I wrote like I would talk.

Speaking out loud, I think ``learnt prediction`` sounds better than ``learned prediction``.
Might just be me :)

python_scripts/02_numerical_pipeline_hands_on.py
lesteve: Given that this part is about numerical variables, we should avoid making remarks on categorical variables that we are going to discard anyway ... I would remove the thing about unequal distribution between categories.

jupyter-book/linear_models/linear_models_module_intro.md
glemaitre: you can just rewrap (single break line). I think Python people don't like backslash :)

python_scripts/02_numerical_pipeline_hands_on.py
glemaitre:
```suggestion
# Let's take for example "education-num".
```

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -93,41 +96,55 @@
...
-# check the number of occurrence of each category to get convinced.
+# Let’s take for example “education-num”.
+# We can assert that each value, which is a number, represents an education
+# level. Therefore, there is only a specific amount of values each sample can
+# take. Let’s print out the distribution to get an intuition of the data:
 
 # %%
-data["education-num"].value_counts().sort_index()
+distribution = data["education-num"].value_counts().sort_index()
+print(distribution)
```
glemaitre, line 127: We don't have to use `print`, we will use the format return by pandas:

```suggestion
distribution
```

python_scripts/02_numerical_pipeline_hands_on.py
glemaitre: From the teaching of yesterday, we might want to use some plotting if the output is already readable (even if I internally like plot ;))

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -93,41 +96,55 @@
...
+plt.bar(distribution.index, distribution)
+plt.title(f"Distribution of 'education-num'")
+plt.xlabel("Category")
+plt.ylabel("Occurrences")
+plt.show()
 
 # %% [markdown]
-# Therefore, we should ignore such type of columns because they would require
-# a specific processing which is different from the continuous variable.
+# We can observe that the distribution is very unequal between categories.
```
glemaitre, line 140:
```suggestion
# We observe that the distribution is very unequal between categories.
```

glemaitre: I prefer to be assertive and avoid `can` whenever possible

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -93,41 +96,55 @@
...
+# Such columns would require a specific processing which is different from
+# continuous values, so we will ignore them for now.
 #
-# Finally, we can mention that we will ignore the `"fnwgt"` column because it
-# corresponds to an hand-crafted variable and we make the choice to only work
-# with variable which has been collected. In the next notebook, we will
-# regularly ignore this variable as well.
+# Finally, because we want to stick to collected data only, we will ignore
```
glemaitre, line 144: collected might not be the best word (I find it not enough explicit) here but I agree that we can go this way to improve the phrasing.

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -141,10 +158,14 @@
 # %% [markdown]
 # ## Train-test split the dataset
 #
-# In the previous notebook, we loaded separately two datasets: a training and a
-# testing dataset. We mentioned that scikit-learn provides an helper function
-# `sklearn.model_selection.train_test_split` allowing to do this split.
-# Here, we will use this tool instead of loading some new data.
+# In the previous notebook, we loaded two separate datasets: a training
+# one and a testing one. However, as mentioned earlier, having separate
+# datasets like that is unusual ; most of the time, we will need to
```
glemaitre, line 163:
```suggestion
# datasets like that are unusual; most of the time, we will need to
```

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -141,10 +158,14 @@
...
+# create them from a single dataset we have.
```
glemaitre, line 164: It is weird to finish a sentence with `we have` :)

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -141,10 +158,14 @@
...
+# \
+# We also mentioned that scikit-learn provides the helper function
+# `sklearn.model_selection.train_test_split` which can be used to
+# automatically make the split.
```
glemaitre, line 168:
```suggestion
# `sklearn.model_selection.train_test_split` which is used to
# automatically split the data.
```

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -23,30 +23,29 @@
...
 
 # %%
 import pandas as pd
 
 df = pd.read_csv("../datasets/adult-census.csv")
 
 # %% [markdown]
-# We will first split our dataset to have the target separated from the data
-# used to train our predictive model.
+# We will now drop the target from the data we will use to train our
```
glemaitre, line 39:
```suggestion
# We will now drop the target from the data that we will use to train our
```

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -126,29 +125,27 @@
...
-#   computation;
-# * predictor internally using gradient-descent based algorithms
-#   (e.g. logistic regression) to find optimal parameters work better
-#   and faster simplify choice of parameters as learning-rate;
-# * predictor using regularization (e.g. logistic regression) require
-#   normalized features such that the penalty is properly applied to the
-#   weights.
+# * predictors using Euclidian distance (e.g k-nearest-neighbors) should have
+#   normalized features so that each one contributes equally to the distance
+#   computation.
```
glemaitre, line 144: I think that we should finish with `;` and not `.` because each item does not start with an upper case. In some way we could have written the same in a single line and the semi-column would have been used.

I would be more inclined to revert these changes everywhere :)

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -171,7 +168,7 @@
 # We can easily combine these sequential operations with a scikit-learn
 # `Pipeline`, which chains together operations and can be used like any other
 # classifier or regressor. The helper function `make_pipeline` will create a
-# `Pipeline` by giving as arguments the successive transformations to perform
+# `Pipeline` ; it takes as arguments the successive transformations to perform,
```
glemaitre, line 171:
```suggestion
# `Pipeline`: it takes as arguments the successive transformations to perform,
```

glemaitre (commented):
A partial review

jupyter-book/predictive_modeling_pipeline/02_numerical_pipeline_quiz.md
```diff
@@ -64,7 +64,7 @@ Cross-validation allows us to:
 +++
 
 ```{admonition} Question
-`make_pipeline` (as well as `Pipeline`) is used for:
+`Pipeline`s are used to:
```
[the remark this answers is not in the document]
Phaide: I won't commit the suggestion since it is a quiz, but you're right !

python_scripts/02_numerical_pipeline_scaling.py
Phaide: Both forms are valid, but I think not using it lightens the sentence, especially on the pronunciation (*the data we* vs *the data that*) :)

python_scripts/02_numerical_pipeline_hands_on.py
Phaide: Rephrase I must

PR #175, opened 2021-01-28
Add javascript file to display content only in iframe

brospars:


lesteve:
Just a question, how much harder would it be to be based on a parameter in the URL, e.g. something like https://inria.github.io/scikit-learn-mooc?content_only=true

IIRC that was the original plan. Also this makes it a bit easier to test. I guess you need to change many more things in FUN to add this parameter where needed ...

Probably a bit tangential but I am wondering whether we want to remove the `<h1>` element as well for some pages (in principle there should be only one `<h1>` on each of our pages right at the beginning). IIRC there was some repetition between the FUN title of the page and the `<h1>` in our HTML in some cases (maybe not the exercise correction but more when we include an HTML page in FUN e.g. for module introduction and take-away) cc @lfarhi

lesteve:
Also technical question although I don't know too much about web things, is there some need to wait for the document to be loaded or something like this before removing some content?

brospars:
> Just a question, how much harder would it be to be based on a parameter in the URL, e.g. something like https://inria.github.io/scikit-learn-mooc?content_only=true
> 
> IIRC that was the original plan. Also this makes it a bit easier to test. I guess you need to change many more things in FUN to add this parameter where needed ...

Yes it was mainly not to have to change urls in FUN but maybe we can do both iframe detection and url parameter ?

> Probably a bit tangential but I am wondering whether we want to remove the `<h1>` element as well for some pages (in principle there should be only one `<h1>` on each of our pages right at the beginning). IIRC there was some repetition between the FUN title of the page and the `<h1>` in our HTML in some cases (maybe not the exercise correction but more when we include an HTML page in FUN e.g. for module introduction and take-away) cc @lfarhi

Yes there was this issue but it was not for iframes so I handle it on the FUN side when the page content is injected

> Also technical question although I don't know too much about web things, is there some need to wait for the document to be loaded or something like this before removing some content?

It's added to the bottom of the page so in theory there is no need to wait

lesteve:
Let's merge this for now I think this is good enough.

PR #177, opened 2021-01-29
Add url param content_only and wait for page to be ready

brospars:
My bad, custom javascript are not added to the bottom of body but to the head... So the script needs to wait for the document to be ready.

I also added the content_only url parameter.


lesteve:
Tested the `content_only` parameter quickly, it seems to work fine, thanks!

PR #250, opened 2021-03-12
MNT simplify plotting whenever possible

### LINKED ISSUE #242: Add additional notes regarding plotting
We sometimes uses advance plotting to have a good rendering.

While some of the style issue could be solved using an rc file (https://github.com/INRIA/scikit-learn-mooc/issues/138) some other issue are more related to the usage of matplotlib or seaborn (`plt.subplots`, etc).

I think that we could:

* add an explanation note to explain what a matplotlib, pandas or seaborn call is doing whenever we think that this specific call will be recurrent and beneficial for the students (e.g. `plt.subplots(ncols=3, ...)`)
* add a note mentioning that the code should not be understood and it is only used for making nice image.

### LINKED ISSUE #243: Add technical notes such that they are better spotted throughout the MOOC
This is a bit of a follow-up to #242 where I mentioned that we could have some notes to explain some technical details.
Up to know, we add `tip` for such thing (e.g. `random_state` and `_ = ...` pattern). However, we also use `tip` boxes for other part of the course (e.g. why one should scale data).

I think that we could be consistent and create a specific technical note that could be easily spotted.

glemaitre:
Simplify plotting whenever possible:

* favor matplotlib or the pandas plotting function that uses matplotlib
* avoid seaborn
* avoid multi plots

closes #243
closes #242


python_scripts/cross_validation_learning_curve.py
```diff
@@ -93,9 +93,8 @@ def make_cv_analysis(regressor, data, target):
 
 # %%
 import matplotlib.pyplot as plt
-import seaborn as sns
 
-sns.displot(scores_sample_sizes, kind="kde")
+scores_sample_sizes.plot.kde()
```
ogrisel, line 97: I wonder if we should refrain from using kernel density estimation plots and instead use histograms.

- con for KDE plots: it is quite a complex concept, our audience does not necessarily what "kde" stands for and might find it too magic and therefore frustrating and we do not want to digress to explain what it is. 
- pro for using KDE plots: the figures is more beautiful (less crowded than an histogram)
- (minor) pro for using KDE plots: they come with automatically selected kernel size by default instead of having to tweak the number of bins to avoid artifacts (although 10 to 30 bins is usually safe)
- pro for using KDE plots: our not-completely-beginners readers might discover that it exists without too much cognitive overload and that it's a nice alternative to histograms and that adds value to our teaching.


Any opinion @GaelVaroquaux in particular?

python_scripts/cross_validation_validation_curve.py
```diff
@@ -138,10 +130,28 @@
 # testing error is minimal, and this is what really matters. This is the
 # best compromise we could reach by just tuning this parameter.
 #
+# Be aware that looking at the mean errors is quite limiting. We should also
+# look at the standard deviation to assess the dispersion of the score. We
+# can repeat the same plot as before but this time, we will had some
```
ogrisel, line 135:
```suggestion
# can repeat the same plot as before but this time, we will add some
```

python_scripts/cross_validation_validation_curve.py
```diff
@@ -138,10 +130,28 @@
...
+# information to show the standard deviation of the errors as well.
+
+# %%
+plt.errorbar(max_depth, train_errors.mean(axis=1),
+             yerr=3 * train_errors.std(axis=1), label='Training error')
+plt.errorbar(max_depth, test_errors.mean(axis=1),
+             yerr=3 * test_errors.std(axis=1), label='Testing error')
```
ogrisel, line 142: Maybe just use 1 std for the error bars. 3 is kind of arbitrary and does not reflect what is written in the paragraph above.

python_scripts/cross_validation_learning_curve.py
glemaitre: For this particular plot, we need the KDE. The histogram will not work. However, while teaching this part, I thought it would be almost unnecessary to have it in the course (maybe move it as an exercise instead). We are using the `sklearn.model_selection.learning_curve` in the subsequent section and this is good enough I think.

python_scripts/cross_validation_validation_curve.py
glemaitre: The error bar will be too small then. Maybe updating the test would be enough?

python_scripts/cross_validation_learning_curve.py
ogrisel: I don't really follow. Can you please clarify what is not necessary? Manually creating the learning curves by hand with the `make_cv_analysis` and directly introducing the concepts with the `learning_curve` function directly?

+1 if so.

python_scripts/cross_validation_validation_curve.py
ogrisel: Why too small? If the standard deviations are small, they are small, no?

python_scripts/cross_validation_learning_curve.py
glemaitre: > Manually creating the learning curves by hand with the make_cv_analysis and directly introducing the concepts with the learning_curve function directly?

Yes exactly. And optionally have `make_cv_analysis` as an exercise if it fits.

ogrisel (commented):
Some comments

ogrisel:
I was writing some comments but you have iterated in the mean time. Feel free to take them into account if they still match you plans.

python_scripts/logistic_regression_non_linear.py
```diff
@@ -46,87 +46,149 @@ def plot_decision_function(fitted_classifier, range_features, ax=None):
...
 
 # %% [markdown]
-# In this later case, we can see that the accuracy is close to be perfect and
-# that the decision boundary is non-linear. Thus, kernel trick or data
-# augmentation are the tricks to make a linear classifier more expressive.
+# We observe something similar than in the previous case. The decision function
+# is more flexible and would be able the generalize.
+#
+#Thus, kernel trick or data augmentation are the tricks to make a linear
+# classifier more expressive, exactly as we saw in regression.
```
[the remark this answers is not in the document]
glemaitre: good point. I will also correct in another place that I wrongly call this data augmentation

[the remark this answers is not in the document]
glemaitre: Actually, we did correct it :)

PR #254, opened 2021-03-17
ENH add a glossary with some terminology

### LINKED ISSUE #38: Glossary?
Would it be useful to have a glossary. Maybe it could help with reminding people of terms and the terminology we use, e.g.,

here we state that we will use the terms 'hyper-parameter' and 'parameter' interchangably:

https://github.com/INRIA/scikit-learn-mooc/blob/adce6e0a9393be7d6e22fc393cdbdaf5698dda68/python_scripts/04_basic_parameters_tuning.py#L23

but this kind of thing may be easy to forget as you work through the material.

glemaitre:
closes #38 

First attempt to add a glossary

### DIFF
--- a/jupyter-book/_toc.yml
+++ b/jupyter-book/_toc.yml
@@ -228,7 +228,7 @@
 
 - part: Appendix
   chapters:
-    - file: appending/glossary
+    - file: appendix/glossary
     - file: appendix/datasets_intro
       sections:
         - file: python_scripts/trees_dataset
--- a/jupyter-book/_toc.yml
+++ b/jupyter-book/_toc.yml
@@ -228,6 +228,7 @@
 
 - part: Appendix
   chapters:
+    - file: appending/glossary
     - file: appendix/datasets_intro
       sections:
         - file: python_scripts/trees_dataset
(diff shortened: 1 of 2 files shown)


lesteve:
I am wondering whether this would be easier to do a CodiMD pad for this one that everyone can edit concurrently (not entirely sure ...)

The alternative is to commit early with a list of terms and TODOs and people do iterative passes when they have 15 minutes or so to improve it.

lesteve:
We can start using https://hackmd.io/@lesteve/scikit-learn-mooc-glossary try to add stuff and copy it back in this PR and merge it early when we think it is good enough.

It does not have to be perfect I would say. Maybe we can leave it open 1-2 days in case we manage to get other people involvement.

glemaitre:
> We can start using https://hackmd.io/@lesteve/scikit-learn-mooc-glossary try to add stuff and copy it back in this PR and merge it early when we think it is good enough.

Looks like a plan

lesteve:
Also finding a good external glossary could be a good idea. Googling a bit, the one I found are slightly too advanced but this one seems good for example:
https://ml-cheatsheet.readthedocs.io/en/latest/glossary.html

lesteve:
Also to be honest this is the kind of thing that can be improved a lot based on feed-back from the beta (typically questions from confused users in the forum).

lesteve:
I am going to merge this one, still some TODOs left but at least this way we can have this in the MOOC.

PR #255, opened 2021-03-17
Add simple numerical data exercise.

lesteve:
This is my attempt at a simple numerical data with k-neighbors classifier.

Note: I originally wanted to do the `n_neighbors=1` 100% accuracy on the train set but not on the test set. Unfortunately life is always more complicated than you think 😢. Because there are many duplicated rows (in this numerical-only dataset, we have 4 features `age`, `hours-per-week`, `capital-gain`, `capital-loss`) with different classes the accuracy is not 100% on the train set.

Not only it does not overfit with `n_neighbors=1` but it actually even underfits (it performs about the same as a majority class predictor both on the train and test set)  for some reason that I let as an exercise for the astute reader.

### DIFF
--- a/python_scripts/02_numerical_pipeline_sol_00.py
+++ b/python_scripts/02_numerical_pipeline_sol_00.py
@@ -42,12 +42,21 @@ target = adult_census["class"]
 # website](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html)
 
 # %% [markdown]
+# The default value for `n_neighbors` is 5
+
+# %% [markdown]
 # Create a `KNeighborsClassifier` model with `n_neighbors=50`
 
 # %%
 from sklearn.neighbors import KNeighborsClassifier
 
 model = KNeighborsClassifier(n_neighbors=50)
+
+# %% [markdown]
+# Fit this model on the data and target loaded above
+
+# %%
+model.fit(data, target)
 
 # %% [markdown]
 # Use your model to make predictions on the first 10 data points inside the
--- a/python_scripts/02_numerical_pipeline_sol_00.py
+++ b/python_scripts/02_numerical_pipeline_sol_00.py
@@ -1,0 +1,92 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.6.0
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
+# %% [markdown]
+# # 📃 Solution for Exercise 00
+#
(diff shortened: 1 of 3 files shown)


python_scripts/02_numerical_pipeline_sol_00.py
```diff
@@ -0,0 +1,101 @@
...
+first_target_values = target.iloc[:10]
+first_target_values
+
+# %%
+number_of_correct_predictions = (
+    first_predictions == first_target_values).sum()
+number_of_predictions = len(first_predictions)
+print(
+    f"{number_of_correct_predictions}/{number_of_predictions} "
+    "of predictions are correct")
```
glemaitre, line 80:
```suggestion
print(f"{number_of_correct_predictions}/"f
      f"{number_of_predictions} of predictions are correct")
```

python_scripts/02_numerical_pipeline_sol_00.py
```diff
@@ -0,0 +1,101 @@
...
+# data. Do they match the actual target values?
+
+# %%
+first_data_values = data.iloc[:10]
+first_predictions = model.predict(first_data_values)
+first_predictions
+
+# %%
+first_target_values = target.iloc[:10]
+first_target_values
```
glemaitre, line 72: To avoid the issue of the rendering (Series vs. array), we can use a `for` loop a bit similar to:

```python
for sample_idx in range(5):
    print(f"Sample #{sample_idx} - true: {target[samples_idx]} vs. "
          f"predicted: {predicted_target[sample_idx}")
```

glemaitre (approved):
LGTM

python_scripts/02_numerical_pipeline_sol_00.py
lesteve: Interesting I slightly prefer my way I don't really know why ...

python_scripts/02_numerical_pipeline_sol_00.py
lesteve: I think good enough for now, you are right that this is not such a big deal ...

lesteve:
OK let's merge this one, thanks for having a look!

python_scripts/02_numerical_pipeline_sol_00.py
glemaitre: Maybe because it is closer to black :)
Throughout the notebooks, the style is to use something the most compact. But this is nitpicking.

lfarhi:
Can you generate the associated notebook 02_numerical_pipeline_ex_00.ipynb ?

lesteve:
Good point, I'll do this.

lesteve:
Done in 5ca67435cfba71b7527437b996e40382872c8892

lfarhi:
Thanks

PR #267, opened 2021-03-23
Use loguniform rather than reciprocal.

lesteve:
loguniform has been added in scipy 1.14 (released December 2019)

log-uniform is easier to grasp (and remember) than reciprocal also this avoids mentioning both names.


GaelVaroquaux:
Thanks!!!

PR #273, opened 2021-04-07
Hide hypothesis elements in content_only mode

### LINKED ISSUE #272: Remove hypothesis side-panel with content_only parameter
https://inria.github.io/scikit-learn-mooc/appendix/glossary.html?content_only

image

On the right-hand side the hypothesis panel is visible. This is also visible in the FUN platform which probably we don't want. This affects also exercise solutions.

Not sure how easy this is to remove with javascript ...

cc @brospars

brospars:
This should resolve #272


lesteve:
Seems like it is working:
https://1399-246063957-gh.circle-artifacts.com/0/jupyter-book/appendix/glossary.html?content_only

Thanks a lot for this!

PR #280, opened 2021-04-23
FIX remove fnwgt for pedagogical reason

### LINKED ISSUE #278: Remove fnlwgt from CSV
This is the simplest. It makes slightly inconsistent with the OpenML dataset but easier not to have to explain why we drop this column (we don't really know actually how it was constructed)

glemaitre:
closes #278


lesteve:
The quizzes needs to be changed in gitlab ...

glemaitre:
Already done and synchronized: https://gitlab.inria.fr/learninglab/mooc-scikit-learn/mooc-scikit-learn-coordination/-/merge_requests/12

PR #281, opened 2021-04-23
FIX always drop education-num and keep quite with this detail

### LINKED ISSUE #279: Remove education-num discussion from the Predictive Modeling Pipeline
The idea is to:
- say that education and education-num contains the same info so that we can drop education-num
- drop education-num at the beginning of each notebook.
- remove the discussion about education-num in the numerical variable notebook.

There may be some quizzes affected by this (needs to double-check)

glemaitre:
closes #279


glemaitre:
ping @lesteve

PR #285, opened 2021-04-27
Move regularized linear models slides to the right sequence

lesteve:
@ogrisel has split the linear models slides in two. This is moving the regularized linear models to their own sequence below.


lesteve:
Note to myself: move the quizz part as well (one question about regularized models)

ogrisel:
Thanks, it looks good to me once the quiz is adapted. I think there is one question on regularized linear models in the first quiz of the linear models module that we be required to a later quiz.

ogrisel (approved):

PR #286, opened 2021-04-30
FIX add a quick explanation of a kNN

glemaitre:
closes https://mooc-forums.inria.fr/moocsl/t/add-nearest-neighbors-explanation-when-we-introduce-it/441

### DIFF
--- a/python_scripts/02_numerical_pipeline_introduction.py
+++ b/python_scripts/02_numerical_pipeline_introduction.py
@@ -79,10 +79,9 @@ print(f"The dataset contains {data.shape[0]} samples and "
 # ## Fit a model and make predictions
 #
 # We will build a classification model using the "K-nearest neighbors"
-# strategy. A K-nearest neighors is a naive model: to predict the target for a
-# new sample, the model will look at the target of the `k` nearest samples
-# of this sample provided during training, and will output the most likely
-# target.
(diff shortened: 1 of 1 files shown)


python_scripts/02_numerical_pipeline_introduction.py
```diff
@@ -79,7 +79,12 @@
 # ## Fit a model and make predictions
 #
 # We will build a classification model using the "K-nearest neighbors"
-# strategy. The `fit` method is called to train the model from the input
+# strategy. A K-nearest neighors is a naive model: to predict the target for a
```
lesteve, line 82:
```suggestion
# To predict the target of a new sample, a K-nearest neighbors takes into
# account its `k` nearest neighbors in the training set and predicts the
# majority target of these `k` nearest neighbors.
```

python_scripts/02_numerical_pipeline_introduction.py
```diff
@@ -79,7 +79,12 @@
...
+# new sample, the model will look at the target of the `k` nearest samples
+# of this sample provided during training, and will output the most likely
+# target.
```
lesteve, line 85:
```suggestion
# strategy. to predict the target of a new sample, a k-nearest neighbors takes
# into account its `k` nearest neighbors in the training set and predicts the
# majority target of these `k` nearest neighbors.
```

PR #287, opened 2021-04-30
ENH add manual figure with code

### LINKED ISSUE #263: Clever way to have hidden code in both JupyterBook and Jupyter Notebook classic and JupyterLab?
Mostly useful for places where we only want to show a plot without showing the code e.g. in the data exploration notebook where we show a 2d plot of a decision tree decision boundaries. Currently we have a script that generates a figure and after this we include the figure in a markdown here like [here](https://github.com/INRIA/scikit-learn-mooc/blob/2d67b1c97d02c2eabe839e5c395c8936be2f6bc4/python_scripts/01_tabular_data_exploration.py#L252). This is not great.

Things to look at:
- a way to have collapsed cells on startup for both JupyterLab (according to [this](https://discourse.jupyter.org/t/hiding-code-cell-on-launch/1763/2) it seems `"source_hidden": true` in the cell metadata should work) and Jupyter Notebook classic (seems to be part of jupyter_contrib_nbextensions see [this](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/hide_input/readme.html) based on `cell.metadata.hide_input`))
- do it the JupyterBook way (see [this](https://jupyterbook.org/interactive/hiding.html#hide-cell-inputs) in the .py file)
- adapt https://github.com/INRIA/scikit-learn-mooc/blob/master/build_tools/convert-python-script-to-notebook.py to have the necessary logic to transform the JupyterBook way in a JupyterLab and Jupyter Notebook classic compatible way.

### LINKED ISSUE #283: Consider getting rid of the decision tree plot in the data exploration notebook
this is a bit hard to explain and this feels out of place compared to the rest of the notebok which is about data exploration.

I think we said that this would make more sense to show a visual hand-made rule with matplotlib vline hline and have a plot that shows:
- left -> low income
- bottom right -> low income
- top right -> ??? (i.e. we don't know)

also mention in one sentence that a decision tree would give simlar rules and help in the top-right cadrant (I think we already say this a bit)

glemaitre:
closes #283 
closes https://mooc-forums.inria.fr/moocsl/t/use-visual-plot-of-hand-made-decision-rules-rather-than-decision-tree-plot/509/4


lesteve:
I am trying to merge this section with the one above that discuss very similar things ...

glemaitre:
OK let me know if I can help reviewing

python_scripts/01_tabular_data_exploration.py
```diff
@@ -224,81 +224,70 @@
...
-# * In the region `age > 28.5 AND hours-per-week < 40.5`
-#   (bottom-right region), the prediction is `low-income`. Note that the blue
-#   is a bit lighter that for the left region which means that the algorithm is
-#   not as certain in this region.
-# * In the region `age > 28.5 AND hours-per-week > 40.5` (top-right region),
-#   the prediction is `low-income`. However the probability of the class
-#   `low-income` is very close to 0.5 which means the model is not sure at all
-#   about its prediction.
+# It is interesting to note that some machine learning models will work
+# similarly to what we did: they are known as decision tree models. Note that
```
lesteve, line 283: It would be nice here to mention that the threshold chosen by hand are arbitrary and that a decision tree will chose the "best" splits based on data.

Also just curious why you switched to 27 years (rather than 25 years). It would be good to mention that these are done by eye and very approximate.

glemaitre: 25 was letting a bunch of blue point on its right (on the figure). 27 corresponds to  the age were the orange dot start (we can actually mention how we decided this rule). I will make the change

glemaitre: I added that this is "one" arbitrary possibility chosen such that we isolate best the two classes with only 2 rules.

PR #325, opened 2021-05-05
FIX be more consistent with regression metrics used

glemaitre:
closes https://mooc-forums.inria.fr/moocsl/t/avoid-switching-between-different-metrics-if-possible/552
closes https://mooc-forums.inria.fr/moocsl/t/explain-r2-score-the-first-time-we-introduce-it/536/2
closes https://mooc-forums.inria.fr/moocsl/t/use-mean-absolute-error-in-the-exercise-as-in-previous-notebook/534


glemaitre:
I am only using the`mean_squared_error` and `mean_absolute_error` and added an introduction from the start.
I think this is still useful to introduce the MSE since this is the error minimize by the linear model (and it is stated in the notebook). Then, I introduce the MAE as a metric that is more intuitive. I use mainly MAE apart when MSE is more relevant (in the regularization notebook for instance).

I remove all other occurrences of $R^2$

PR #328, opened 2021-05-06
Show failure logs after jupyter-book detected failure.

lesteve:
Follow-up for fixing master which was red. This would be a lot better to show the errors directly in the build log rather than having to remember where to look in the artifacts.


lesteve:
Seems to work fine according to https://circleci.com/api/v1.1/project/github/INRIA/scikit-learn-mooc/1814/output/105/0?file=true&allocation-id=60938b68e704a11b9d1a5078-0-build%2F2F597959 merging

PR #353, opened 2021-05-26
Improve randomized search cv notebook

ogrisel:
This PR changes 2 things:

- first, only use the inner parallelism of HistGradientBoosting instead of process level parallism with `n_jobs=4` in the `RandomizedSearchCV` call itself. I noticed it caused oversubsubscription performance problems with our current configuration of `OMP_NUM_THREADS` and the jupyterhub docker spawner config
- secondly, be more explicit about how to analyze the plot.

The second point is based on the feedback of:

https://mooc-forums.inria.fr/moocsl/t/video-to-explain-how-to-use-the-interactive-parallel-coordinates-plot/2548

I know this is kind of partially answering the quiz but I think it is fine. The goal of the quiz is check that the student carefully read and understood the analysis and could reproduce it and refine it further.

I think having a video to redo this analysis would be a nice addition but this PR is already a net improvement I think.

### DIFF
--- a/notebooks/parameter_tuning_randomized_search.ipynb
+++ b/notebooks/parameter_tuning_randomized_search.ipynb
@@ -145,8 +145,9 @@
     "\n",
     "model = Pipeline([\n",
     "    (\"preprocessor\", preprocessor),\n",
-    "    (\"classifier\",\n",
-    "     HistGradientBoostingClassifier(random_state=42, max_leaf_nodes=4))])\n",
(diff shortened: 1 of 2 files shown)


glemaitre:
> I think having a video to redo this analysis would be a nice addition but this PR is already a net improvement I think.

In the course notebook (before the exercise), I was thinking that to have at least an animated GIF regarding the interaction with the parallel coordinates plot

glemaitre (approved):
LGTM.

I just let the CI finish

PR #359, opened 2021-06-01
Fix typo in parameter_tuning_randomized_search.py

pierreloicq:
Since we want to usE this distribution to create integer


lesteve:
Thanks a lot for the PR, merging!

PR #362, opened 2021-06-07
Pin jupyter-book to 0.10.2.

lesteve:
0.11 is a breaking change that changes the format of _toc.yml see https://github.com/executablebooks/jupyter-book/blob/master/CHANGELOG.md#v0111.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,6 +8,6 @@ dependencies:
   - matplotlib-base
   - seaborn
   - plotly
-  - jupyter-book
+  - jupyter-book=0.10.2
   - jupytext
   - beautifulsoup4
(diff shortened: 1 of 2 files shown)


lesteve:
This is the reason for the failure on master (see https://github.com/INRIA/scikit-learn-mooc/runs/2764651719):
```
===============================================================================

The Table of Contents file is malformed: toc is not a mapping: <class 'list'>
You may need to migrate from the old format, using:

	jupyter-book toc migrate /home/runner/work/scikit-learn-mooc/scikit-learn-mooc/jupyter-book/_toc.yml -o /home/runner/work/scikit-learn-mooc/scikit-learn-mooc/jupyter-book/_toc.yml

===============================================================================
```

PR #363, opened 2021-06-07
Update _toc.yml after jupyter-book 0.11

lesteve:
Ran `jupyter-book toc migrate _toc.yml -o _toc.yml` and added back some emoticons 🚧 which had been converted by the migration tool.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,6 +8,6 @@ dependencies:
   - matplotlib-base
   - seaborn
   - plotly
-  - jupyter-book
+  - jupyter-book >= 0.11
   - jupytext
   - beautifulsoup4
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -8,6 +8,6 @@ dependencies:
   - matplotlib-base
   - seaborn
   - plotly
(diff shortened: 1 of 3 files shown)


lesteve:
Need to compare:

### this PR
https://1979-246063957-gh.circle-artifacts.com/0/jupyter-book/index.html

### master
https://inria.github.io/scikit-learn-mooc/

lesteve:
Merging this one.

- I looked at the websites they seem visually identical
- I looked at a few page HTML diffs, and from what I can see only the left-hand-side panel has changed (as expected from the jupyter-book changelog https://github.com/executablebooks/jupyter-book/blob/master/CHANGELOG.md?rgh-link-date=2021-06-07T13%3A57%3A26Z#v0111).
- I checked that adding `?content_only` still works fine and double-checked that https://github.com/INRIA/scikit-learn-mooc/blob/master/jupyter-book/_static/sklearn_mooc.js is not using any of the classes that have changed in the LHS panel
- given that the LHS panel is not used in FUN and the previous bullet point, this is unlikely to affect the MOOC in the FUN platform

Of course if you notice unwanted side-effects let me know!

PR #372, opened 2021-07-02
Mygoodbranch

ArturoAmorQ:


python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -302,11 +302,12 @@
...
-# testing set. This strategy has several issues: in a setting where the amount
-# of data is small, the subset used to train or test will be small. Besides, a
-# single split does not give information regarding the confidence of the
-# results obtained.
+# In the previous example, we split the original data into a training set and a 
+# testing set. The score of a model will in general depend on the way we make 
+# such split. Because of this, a single split does not give full information 
+# regarding the confidence of the results obtained. Besides, in a setting where 
+# the amount of data is small, the data available for training and testing will 
+# be even smaller after splitting.
```
ArturoAmorQ, line 310: The idea is to emphasize that using a simple train/test split is not wrong, but could be improved.

PR #373, opened 2021-07-02
Improve single split wording

ArturoAmorQ:
The idea is to emphasize that using a simple train/test split is not wrong, but could be improved.

### DIFF
--- a/python_scripts/02_numerical_pipeline_scaling.py
+++ b/python_scripts/02_numerical_pipeline_scaling.py
@@ -304,10 +304,10 @@ print(f"The accuracy using a {model_name} is {score:.3f} "
 #
 # In the previous example, we split the original data into a training set and a 
 # testing set. The score of a model will in general depend on the way we make 
-# such split. Because of this, a single split does not give full information 
-# regarding the confidence of the results obtained. Besides, in a setting where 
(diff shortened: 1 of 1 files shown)


python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -302,11 +302,12 @@
...
 # ## Model evaluation using cross-validation
 #
-# In the previous example, we split the original data into a training set and a
-# testing set. This strategy has several issues: in a setting where the amount
-# of data is small, the subset used to train or test will be small. Besides, a
-# single split does not give information regarding the confidence of the
-# results obtained.
+# In the previous example, we split the original data into a training set and a 
+# testing set. The score of a model will in general depend on the way we make 
+# such a split. One downside of doing a single split is that it does not give
```
ArturoAmorQ, line 307:
```suggestion
# such a split. Doing a single split does not give
```

ArturoAmorQ (commented):
Just another suggestion

python_scripts/02_numerical_pipeline_scaling.py
lesteve: Hmmm I think it is good enough as it is but thanks for the suggestion 😉

PR #374, opened 2021-07-03
Test cache with PR from same repo

lesteve:


lesteve:
OK it works with a PR from the same repo ...

PR #378, opened 2021-07-05
FIX grammar

ArturoAmorQ:
See link below for rules on the use of a/an

https://forum.wordreference.com/threads/a-an-unitary.42391/


lesteve:
Thanks merging this one!

PR #381, opened 2021-07-06
Feature/my quiz

Enriquesec:


Enriquesec:
My solutions

lesteve:
I am guessing that you intended to create a PR in your fork and not in this repo, closing this one!

By the way it would be great if you can avoid putting quiz solutions in a public repo, otherwise people may find it and their MOOC quiz scores will not mean anything ...

PR #393, opened 2021-07-12
Fix typos and wording

ArturoAmorQ:
PR includes the .ipynb files already built

PR #394, opened 2021-07-12
Rephrase and add hint

ArturoAmorQ:
Rephrased for clarity, added hint and built the .ipynb file


notebooks/cross_validation_ex_01.ipynb
```diff
@@ -105,11 +105,11 @@
     "As previously mentioned, the parameter `gamma` is one of the parameter\n",
     "controlling under/over-fitting in support vector machine with an RBF kernel.\n",
     "\n",
-    "Compute the validation curve\n",
-    "(using [`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html))\n",
-    "to evaluate the effect of the parameter `gamma`. You can vary its value\n",
-    "between `10e-3` and `10e2` by generating samples on a logarithmic scale.\n",
-    "Thus, you can use `np.logspace(-3, 2, num=30)`.\n",
+    "Evaluate the effect of the parameter `gamma` by using a\n",
+    "[`validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html)).\n",
```
ogrisel, line 109: I think it's helpful to put the full module path in the question to make it easier to know where to import it from.

I know, the student could just look up the import statement from the previous notebook but since switching between notebooks is slow in Fun, let's be helpful here.

```suggestion
    "[`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html)).\n",
```

ogrisel (commented):
I started to review and just realized that you edited the `.ipynb` file. Please edit the Python script instead instead. This makes the diff much more easier to read.

We regenerate the notebooks from the python scripts from time to time using the `make` command (see the top level `Makefile`).

python_scripts/cross_validation_ex_01.py
```diff
@@ -65,11 +65,11 @@
 # As previously mentioned, the parameter `gamma` is one of the parameter
 # controlling under/over-fitting in support vector machine with an RBF kernel.
 #
-# Compute the validation curve
-# (using [`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html))
-# to evaluate the effect of the parameter `gamma`. You can vary its value
-# between `10e-3` and `10e2` by generating samples on a logarithmic scale.
-# Thus, you can use `np.logspace(-3, 2, num=30)`.
+# Evaluate the effect of the parameter `gamma` by using a
+# [`validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html)).
```
ogrisel, line 69: I think it's helpful to put the full module path in the question to make it easier to know where to import it from.

I know, the student could just look up the import statement from the previous notebook but since switching between notebooks is slow in Fun, let's be helpful here.

```suggestion
# [`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html)).
```

ogrisel (approved):
Besides the previous comment, LGTM!

lesteve:
This is slightly painful, but in principle when you change `some_file_ex_some_number.py`, you should change `some_file_sol_some_number.py`.

The instructions of the exercise are copied and pasted into the solution.

lesteve:
General tip: `git grep` is quite useful for this kind of things where you changed something and you want to make sure there are no other places where the same problem happens:

```
❯ git grep -i 'Compute the validation'
notebooks/cross_validation_ex_01.ipynb:    "Compute the validation curve\n",
notebooks/cross_validation_sol_01.ipynb:    "Compute the validation curve to evaluate the effect of the parameter `gamma`.\n",
python_scripts/cross_validation_ex_01.py:# Compute the validation curve
python_scripts/cross_validation_sol_01.py:# Compute the validation curve to evaluate the effect of the parameter `gamma`.
```

You can also remove the `.ipynb` files from the output like this:
```
❯ git grep -i 'Compute the validation' | grep -v '.ipynb'
python_scripts/cross_validation_ex_01.py:# Compute the validation curve
python_scripts/cross_validation_sol_01.py:# Compute the validation curve to evaluate the effect of the parameter `gamma`.
```

As you can see the `cross_validation_ex_01.py` and `cross_validation_sol_01.py` contents have already started to drift away in this case ...

ogrisel:
Indeed, I merged the previous PR too quickly: https://github.com/INRIA/scikit-learn-mooc/pull/393/files#diff-8c67e2a9e25ae56071b0f6dfaaadc768abb8bec5af032ee247658d962ecee3f2 needs to also be reported in the solution file.

lesteve:
To be honest, maybe there is something we can do to auto-generate the `ex` file from the `sol` file, in a similar way we generate the quizzes without answers from the quizz with answers from the gitlab (i.e. using cell tags, see https://github.com/INRIA/scikit-learn-mooc/blob/master/jupyter-book/convert_exercises.py)

A bit hacky but maybe hacky in a useful way?

python_scripts/cross_validation_ex_01.py
```diff
@@ -65,11 +65,11 @@
 # As previously mentioned, the parameter `gamma` is one of the parameter
 # controlling under/over-fitting in support vector machine with an RBF kernel.
 #
-# Compute the validation curve
-# (using [`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html))
-# to evaluate the effect of the parameter `gamma`. You can vary its value
-# between `10e-3` and `10e2` by generating samples on a logarithmic scale.
-# Thus, you can use `np.logspace(-3, 2, num=30)`.
+# Evaluate the effect of the parameter `gamma` by using a
+# [`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html)).
```
ogrisel, line 69:
```suggestion
# Evaluate the effect of the parameter `gamma` by using the
# [`sklearn.model_selection.validation_curve`](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.validation_curve.html))
# function.
```

python_scripts/cross_validation_ex_01.py
```diff
@@ -65,11 +65,11 @@
...
+# Use the default `scoring=None`. You can vary `gamma`
```
ogrisel, line 70:
```suggestion
# You can leave the default `scoring=None` which is equivalent to
# `scoring="accuracy"` for classification problems. You can vary `gamma`
```

PR #397, opened 2021-07-13
FIX edit solution files

ArturoAmorQ:
I edited the ex files but forgot to make the same modifications to the sol files. This fixes it.

### DIFF
--- a/python_scripts/cross_validation_sol_01.py
+++ b/python_scripts/cross_validation_sol_01.py
@@ -9,8 +9,8 @@
 #   cross-validation;
 # * study the effect of the parameter gamma of this classifier using a
 #   validation curve;
-# * study if it would be useful in term of classification if we could add new
-#   samples in the dataset using a learning curve.
+# * use a learning curve to determine the usefulness of adding new
+#   samples in the dataset when building a classifier.
 #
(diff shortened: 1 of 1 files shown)


ogrisel (approved):

python_scripts/cross_validation_sol_01.py
```diff
@@ -80,16 +80,18 @@
...
 
 # %% [markdown]
-# As previously mentioned, the parameter `gamma` is one of the parameter
+# As previously mentioned, the parameter `gamma` is one of the parameters
 # controlling under/over-fitting in support vector machine with an RBF kernel.
 #
-# Compute the validation curve to evaluate the effect of the parameter `gamma`.
-# You can vary its value between `10e-3` and `10e2` by generating samples on a
-# logarithmic scale. Thus, you can use `np.logspace(-3, 2, num=30)`.
+# "Evaluate the effect of the parameter `gamma` by using a\n",
```
lesteve, line 86: @ArturoAmorQ something has gone slightly wrong here, maybe you have copied an pasted `"` and `\n`?

You can see the result [here](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_sol_01.html):

image

Issue #376, opened 2021-07-05
Emphasize that one should not use OneHotEncoder with tree-based models


glemaitre:
I see a recurrent topic where people still want to use a `OneHotEncoder` with tree-based models in the tree module of the MOOC.
Indeed, I thought that we emphasize that one should not do that such as in this [example](https://inria.github.io/scikit-learn-mooc/python_scripts/03_categorical_pipeline_sol_02.html).

But the message is probably not strong enough. Thus, in some exercises with trees, we do not specify the type of encoding and people complain. I am not sure that we should be more explicit in the exercises but we should make sure that using a `OneHotEncoder` should be seen as a bad practice with tree-based models in general.

ogrisel:
Note that using OHE for trees wouldn't be such a bad practice in general.  For instance lightgbm naturally support optimized handling of anti-correlated sparse features (look for "Exclusive Feature Bundling" in the lightgbm paper).

However, in the context of the scikit-learn implementation, it is almost always the case that `OrdinalEncoding` would be significantly faster than fitting a tree on dense OHE encodings.

For `HistGradientBoosting` we should mention the native categorical at some point in the feature, once they are easy to use with feature names.

lesteve:
Maybe some kind of summary would help (not sure where, maybe a lesson-wide take-home message for the categorical data handling?). The simplified version may be something like this (but there may be some additional subtleties that we want to add):

Which encore should I use?

- OneHotEncoder: to will always do something meaningful, but can be unnecessary slow with trees
- OrdinalEncoder: can be detrimental for linear models unless your category has a meaningful order and you make sure that OrdinalEncoder respects this order. Trees can deal with OrdinalEncoder fine as long as they are deep enough.

ArturoAmorQ:
The solution I am currently working on:

- Rephrase text in [03_categorical_pipeline.py](https://github.com/INRIA/scikit-learn-mooc/blob/master/python_scripts/03_categorical_pipeline.py) for clarity and highlight references to exercise notebooks.

- Add table and closing remarks to notebooks [03_categorical_pipeline_ex_02.py](https://github.com/INRIA/scikit-learn-mooc/blob/master/python_scripts/03_categorical_pipeline_ex_02.py) and [03_categorical_pipeline_sol_02.py](https://github.com/INRIA/scikit-learn-mooc/blob/master/python_scripts/03_categorical_pipeline_sol_02.py)


| | Meaningful order | Non-meaningful order
[...]

PR #400, opened 2021-07-19
Add a summary about "which encoder should I use"

ArturoAmorQ:
Made the following changes to address Issue #376

- Rephrased text in [03_categorical_pipeline.py](https://github.com/INRIA/scikit-learn-mooc/blob/master/python_scripts/03_categorical_pipeline.py) for clarity and highlight references to exercise notebooks.

- Added table and closing remarks to notebooks

| | Meaningful order | Non-meaningful order
------------ | ------------ | ------------- 
 Tree-based model | `OrdinalEncoder` | `OrdinalEncoder`
 Linear model | `OrdinalEncoder` with caution | `OneHotEncoder`


python_scripts/03_categorical_pipeline_ex_02.py
```diff
@@ -101,3 +101,20 @@
 
 # %%
 # Write your code here.
+
+# %% [markdown]
```
lesteve, line 105: Probably this summary table should only be present in the solution and not the exercise.

In other words the exercise is an example which gives intuitions and kind of builds up towards the summary table.

python_scripts/03_categorical_pipeline_sol_02.py
```diff
@@ -153,3 +153,20 @@
 # The main take away message is that arbitrary integer coding of
 # categories is perfectly fine for `HistGradientBoostingClassifier`
 # and yields fast training times.
+
+# %% [markdown]
```
lesteve, line 157: After the work I did on generating the exercise from the solutions in #408, you need to use `# %% [markdown] tags=["solution"]`, see https://github.com/INRIA/scikit-learn-mooc/blob/master/CONTRIBUTING.md#generating-exercises-from-the-solutions for more details.

Also make sure that doing `make exercises` is not modifying your particular exercise (It will modify other exercises which we need to fix in #418

python_scripts/03_categorical_pipeline.py
```diff
@@ -247,28 +247,34 @@
 #
 # Choosing an encoding strategy will depend on the underlying models and the
 # type of categories (i.e. ordinal vs. nominal).
+
+# %% [markdown]
+# ```{note}
+# In general `OneHotEncoder` is the encoding strategy used when the
+# downstream models are **linear models** while `OrdinalEncoder` is used with
+# **tree-based models**.
```
ogrisel, line 255:
```suggestion
# downstream models are **linear models** while `OrdinalEncoder` is often a
# good strategy with **tree-based models**.
```

ogrisel (approved):
Thanks for the nice improvements! Just a few suggestions/comments:

glemaitre:
Thanks @ArturoAmorQ @ogrisel

PR #401, opened 2021-07-20
Fix wording on grid-search parameter tuning

ArturoAmorQ:
Made several changes to address Issue #376, i.e., to give a clear reminder of the use of `OrdinalEncoder` and `OneHotEncoder`.

Fixed some typos, grammar and markdown.

### DIFF
--- a/python_scripts/parameter_tuning_grid_search.py
+++ b/python_scripts/parameter_tuning_grid_search.py
@@ -49,14 +49,9 @@ data_train, data_test, target_train, target_test = train_test_split(
 # We will define a pipeline as seen in the first module. It will handle both
 # numerical and categorical features.
 #
-# As we will use a tree-based model as a predictor, here we apply an ordinal
-# encoder on the categorical features: it encodes every category with an
-# arbitrary integer. For simple models such as linear models, a one-hot encoder
-# should be preferred. But for complex models, in particular tree-based models,
-# the ordinal encoder is useful as it avoids having high-dimensional
-# representations.
-#
-# First we select all the categorical columns.
+# Here we will use a tree-based model as a predictor. That means that only the
+# categorical features need preprocessing. The first step is to select all the
+# categorical columns.
 
 # %%
 from sklearn.compose import make_column_selector as selector
@@ -65,7 +60,13 @@ categorical_columns_selector = selector(dtype_include=object)
 categorical_columns = categorical_columns_selector(data)
 
 # %% [markdown]
-# Then we build our ordinal encoder, giving it the known categories.
+# Here we apply an `OrdinalEncoder` on the categorical features, which encodes 
+# every category with an arbitrary integer. Remember that for simple models 
+# such as linear models, a `OneHotEncoder` should be preferred. But for complex 
+# models, in particular tree-based models, the `OrdinalEncoder` avoids having 
+# high-dimensional representations.
+#
+# We now build our `OrdinalEncoder`, giving it the known categories.
 
 # %%
 from sklearn.preprocessing import OrdinalEncoder
@@ -74,8 +75,8 @@ categorical_preprocessor = OrdinalEncoder(handle_unknown="use_encoded_value",
                                           unknown_value=-1)
 
 # %% [markdown]
-# We now use a column transformer with code to select the categorical columns
-# and apply to them the ordinal encoder.
+# We then use a `ColumnTransformer` to select the categorical columns and
+# apply the `OrdinalEncoder` to them.
 
 # %%
 from sklearn.compose import ColumnTransformer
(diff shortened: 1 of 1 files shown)


python_scripts/parameter_tuning_grid_search.py
```diff
@@ -245,7 +246,7 @@ def shorten_param(param_name):
 #   increased, one should decrease the value of `learning_rate` accordingly
 #   to preserve a good accuracy.
 #
-# The precise meaning of those two parameters will be explained in a latter
+# The precise meaning of those two parameters will be explained in a later
```
ArturoAmorQ, line 249: See https://www.quora.com/What-is-the-difference-between-later-and-latter

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -49,14 +49,9 @@
...
 #
-# As we will use a tree-based model as a predictor, here we apply an ordinal
-# encoder on the categorical features: it encodes every category with an
-# arbitrary integer. For simple models such as linear models, a one-hot encoder
-# should be preferred. But for complex models, in particular tree-based models,
-# the ordinal encoder is useful as it avoids having high-dimensional
-# representations.
-#
-# First we select all the categorical columns.
+# Here we will use a tree-based model as a predictor. That means that only the
```
lesteve, line 52: Maybe this would be worth simplifying this part and the part below to only say something about tree-based models (and not mention linear models since we are not using them in this notebook):
- they don't need scaling for the numerical variables
- they can deal with an `OrdinalEncoder` even if the integer coding order is not meaningful

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -104,10 +105,10 @@
 # ## Tuning using a grid-search
 #
 # Instead of manually writing the two `for` loops, scikit-learn provides a
-# class called `GridSearchCV` which implement the exhaustive search implemented
+# class called `GridSearchCV` which implements the exhaustive search we did
 # during the exercise.
```
lesteve, line 109: I guess "previous exercise" would help here (this is a pity we don't have ways to create links that work for all our setups, i.e. FUN-MOOC, JupyterBook, Jupyter notebook ...)

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -104,10 +105,10 @@
 # ## Tuning using a grid-search
 #
 # Instead of manually writing the two `for` loops, scikit-learn provides a
-# class called `GridSearchCV` which implement the exhaustive search implemented
+# class called `GridSearchCV` which implements the exhaustive search we did
```
lesteve, line 108: I think this whole sentence can be simplified. Maybe something like this (improvement welcome!):

In the previous exercise, we used multiple `for` loops to find the best hyperparameter values over a fixed grid of parameters. `GridSearchCV` is a scikit-learn class that implements a very similar logic with less boiler-plate code.

lesteve (commented):
A few comments

python_scripts/parameter_tuning_grid_search.py
lesteve: Now that the exercises are generated automatically from the corrections, you will need to use the things introduced in https://github.com/INRIA/scikit-learn-mooc/pull/408#issuecomment-884790196

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -103,11 +104,11 @@
...
 #
-# Instead of manually writing the two `for` loops, scikit-learn provides a
-# class called `GridSearchCV` which implement the exhaustive search implemented
-# during the exercise.
+# In the previous exercise we used one `for` loop for each hyperparameter to find the 
+# best combination over a fixed grid of values. `GridSearchCV` is a scikit-learn class 
+# that implements a very similar logic with less repetitive code.
 #
-# Let see how to use the `GridSearchCV` estimator for doing such search.
+# Lets see how to use the `GridSearchCV` estimator for doing such search.
```
ogrisel, line 111:
```suggestion
# Let's see how to use the `GridSearchCV` estimator for doing such search.
```

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -245,7 +246,7 @@ def shorten_param(param_name):
 #   increased, one should decrease the value of `learning_rate` accordingly
 #   to preserve a good accuracy.
 #
-# The precise meaning of those two parameters will be explained in a latter
+# The precise meaning of those two parameters will be explained in a later
 # notebook.
```
ogrisel, line 250: "latter" was intended (https://www.gingersoftware.com/content/grammar-rules/adjectives/later-vs-latter/) but maybe it's wrong to use "a latter" instead of "the latter".

Maybe instead we can just say:

```suggestion
# The precise meaning of those two parameters will be explained later.
```

ogrisel (approved):
LGTM!

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -65,7 +58,15 @@
 categorical_columns = categorical_columns_selector(data)
 
 # %% [markdown]
-# Then we build our ordinal encoder, giving it the known categories.
+# Here we will use a tree-based model as a predictor. That means:
```
glemaitre, line 61: To avoid `predictor`.

```suggestion
# Here we will use a tree-based model as a classifier
# (i.e. `HistGradientBoostingClassifier`). That means:
```

glemaitre (approved):
Just one small nitpick to avoid `predictor`

glemaitre:
Thanks @ArturoAmorQ

Issue #352, opened 2021-05-26
Better terms for predictor, regressor, classifier throughout the course


lesteve:
This is probably too much jargon, also there is a source for confusion because regressors can also mean features used in a regression (similar thing for predictors) as mentioned in the scikit-learn glossary for example: https://scikit-learn.org/stable/glossary.html#term-predictor
> In statistics, "predictors" refers to features.

The terms we agreed on:
- predictor -> predictive model
- regressor -> regression model
- classifier -> classification model

lesteve:
We probably want to remove the corresponding entries in the [glossary](https://github.com/INRIA/scikit-learn-mooc/blob/master/jupyter-book/appendix/glossary.md).

ArturoAmorQ:
> We probably want to remove or update the corresponding entries in the glossary ...

Just a bit awkward that "Classifier" and "Regressor" would still be embedded in `KNeighborsClassifier` and `DecisionTreeRegressor` but I cannot find any simple work-around in this case.

lesteve:
Good point we may want to keep the regressor and classifier entries in the glossary then ...

Another point, probably not that important, but my feeling was also that "regressor"/"classifier" was quite often used to mean a scikit-learn object as in the glossary definition and that "regression model"/"classification model" has a more general meaning.

ArturoAmorQ:
One option is to keep both terms in the glossary and add a small explanation, for instance

```
### regression model

A regression model is a [predictive model](#predictive model) in a [regression](#regression)
setting.

In scikit-learn, `DecisionTreeRegressor` and `Ridge` are regression model classes

### regressor

The term "regressor" is quite often used to mean a scikit-learn object implementing a [regression model] (#regression model). 

Note:  "regressors" can also mean features used in a regression 
```

Or even just keeping the latter definition may suffice. What do you think?

GaelVaroquaux:
> What do you think?

I like your suggestion above!

lesteve:
The weak consensus from Tuesday meeting is that it was not worth it, too much effort for too little impact.

PR #404, opened 2021-07-21
Replace statistical performance by generalization performance

### LINKED ISSUE #322: Use generalization performance rather than statistical performance
At one point we decided statistical performance was better but now we have second thoughts so maybe for the v2?

The consensus seems to be:
- generalization performance slightly better than
- predictive performance which is better than
- statistical performance (too vague)

ArturoAmorQ:
Fix #322 (Use generalization performance rather than statistical performance)

Special care was taken for [02_numerical_pipeline_introduction.py](https://github.com/INRIA/scikit-learn-mooc/blob/master/python_scripts/02_numerical_pipeline_introduction.py) to keep self-consistency.

### DIFF
--- a/python_scripts/02_numerical_pipeline_ex_01.py
+++ b/python_scripts/02_numerical_pipeline_ex_01.py
@@ -15,7 +15,7 @@
 # %% [markdown]
 # # 📝 Exercise M1.03
 #
-# The goal of this exercise is to compare the statistical performance of our
+# The goal of this exercise is to compare the generalization performance of our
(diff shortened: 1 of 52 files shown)


lesteve:
I edited your PR title to be more meaningful (a meaningful helps when looking at all the PRs and also it becomes the commit message by default when doing squash+merge).

Other than this LGTM, I double-checked that there were no remaining instances of "statistical performance"

lesteve:
@ArturoAmorQ in a very similar spirit you have https://github.com/INRIA/scikit-learn-mooc/issues/352 nudge, nudge 😉

PR #408, opened 2021-07-22
Generate exercises from the solutions and fix problems in first module exercises

### LINKED ISSUE #396: Generate exercise instructions `_ex_N.py` file from solution `_sol_N.py` file
From https://github.com/INRIA/scikit-learn-mooc/pull/394#issuecomment-879179614

> To be honest, maybe there is something we can do to auto-generate the `ex` file from the `sol` file, in a similar way we generate the quizzes without answers from the quizz with answers from the gitlab (i.e. using cell tags, see https://github.com/INRIA/scikit-learn-mooc/blob/master/jupyter-book/convert_exercises.py)

> A bit hacky but maybe hacky in a useful way?

We could use `solution` tag for the cells that need to be included in the correction but not the exercise instructions.

Question: what is the syntax to add cell tag in a `.py` file? Answer: this seems to work and is simple enough
```py
# %% tags=["parameters"]
param = 5
```

This converts into this the following `.md` with jupytext:
``````
```python tags=["parameters"]
param = 5
```
``````

lesteve:
Fix #396.

TODO:
- [x] add entry in the Makefile to conveniently generate exercise from solution
- [x] clean up script a bit
- [x] avoid unnecessary jupytext header changes in the script? Let's not bother about this too much for now (I left a TODO in the script)!
- [x] more? Good enough for now!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,6 @@ copy_matplotlibrc:
 
 sanity_check_$(NOTEBOOKS_DIR):
 	python build_tools/sanity-check.py $(PYTHON_SCRIPTS_DIR) $(NOTEBOOKS_DIR)
+
+exercises:
+  python generate_exercise_from_solution.py $(PYTHON_SCRIPTS_DIR)
--- a/Makefile
+++ b/Makefile
@@ -19,4 +19,4 @@ sanity_check_$(NOTEBOOKS_DIR):
 	python build_tools/sanity-check.py $(PYTHON_SCRIPTS_DIR) $(NOTEBOOKS_DIR)
 
 exercises:
-	python generate_exercise_from_solution.py $(PYTHON_SCRIPTS_DIR)
+	python build_tools/generate-exercise-from-solution.py $(PYTHON_SCRIPTS_DIR)
--- a/Makefile
+++ b/Makefile
@@ -19,4 +19,4 @@ sanity_check_$(NOTEBOOKS_DIR):
 	python build_tools/sanity-check.py $(PYTHON_SCRIPTS_DIR) $(NOTEBOOKS_DIR)
 
 exercises:
-  python generate_exercise_from_solution.py $(PYTHON_SCRIPTS_DIR)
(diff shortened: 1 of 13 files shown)


lesteve:
OK this is why but here is how it works for now:
- using `# solution` comment inside a cell will remove all the content afterwards (including the `# solution` line) and replace it with `# Write your code here.`
- using `tags=["solution"]` removes the whole cell.

Also this shows that there are some changes from @ArturoAmorQ that were only done to exercise and not solutions (this was our fault for not catching it but this is hard to remember all these quirks and we are human ...)

I guess working on this PR will allow us to catch more places where the exercises and the solution has drifted away ...

lesteve:
Looks like we would need some special treatment for solution cells that we want replace with "# Write your code here." and that have a cell magic ... 😢 because of https://github.com/jupyter/notebook/issues/4394#issuecomment-462104611.

Since python_scripts/03_categorical_pipeline_sol_02.py is the only one to use this I am going to use a `time.time()` instead ...

lesteve:
OK I am going to say this is good enough to get some feed-back!

ArturoAmorQ (approved):
LGTM!

glemaitre (approved):
LGTM

glemaitre:
Thanks @lesteve

miykael:
FYI: With regards to **hiding the Jupytext header** in python scripts, I found this issue https://github.com/mwouts/jupytext/issues/527, that points out that the `hide_notebook_metadata=true` flag exists. But combining it with jupytext's `writes()` command [here](https://github.com/INRIA/scikit-learn-mooc/blob/main/build_tools/generate-exercise-from-solution.py#L43) seems to require a `config` file.

lesteve:
Thanks for pointing this out, I looked a bit at this and somehow this seems only applied for markdown file:
https://github.com/mwouts/jupytext/blob/8cfa66bef380bf6d54479a06b092da30dbc9a7c4/jupytext/header.py#L138-L142

I think for now the best is to live with this additional jupytext metadata in exercise files ...

PR #423, opened 2021-08-02
Fix redundant text in linear_regression_without_sklearn.py

ArturoAmorQ:
Minor wording to avoid repeating ourselves


lesteve:
Merging thanks!

PR #422, opened 2021-07-30
Change exercise solutions and regenerate exercise instructions

ArturoAmorQ:
Fixes #418

Solution files to modify:
- [x] python_scripts/cross_validation_sol_01.py  
- [x] python_scripts/cross_validation_sol_02.py 
- [x] python_scripts/ensemble_sol_01.py         
- [x] python_scripts/ensemble_sol_02.py         
- [x] python_scripts/ensemble_sol_03.py         
- [x] python_scripts/ensemble_sol_04.py         
- [x] python_scripts/ensemble_sol_05.py         
- [x] python_scripts/feature_selection_sol_01.py
- [x] python_scripts/linear_models_sol_01.py (Loïc)   
- [x] python_scripts/linear_models_sol_02.py (Loïc)   
- [x] python_scripts/linear_models_sol_03.py (Loïc)    
- [x] python_scripts/linear_models_sol_04.py (Loïc)   
- [x] python_scripts/linear_models_sol_05.py (Loïc)    
- [x] python_scripts/metrics_sol_01.py          
- [x] python_scripts/metrics_sol_02.py          
- [x] python_scripts/parameter_tuning_sol_02.py (Olivier)
- [x] python_scripts/parameter_tuning_sol_03.py (Olivier)
- [x] python_scripts/trees_sol_01.py (Olivier)            
- [x] python_scripts/trees_sol_02.py (Olivier)

PR #424, opened 2021-08-02
Fix regenerate exercise M4.01 from solution

ArturoAmorQ:
PR #422 erased a hint in [exercise M4.01](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_ex_01.html).
This fixes it while improving wording.

### DIFF
--- a/python_scripts/linear_models_sol_01.py
+++ b/python_scripts/linear_models_sol_01.py
@@ -46,10 +46,10 @@ def linear_model_flipper_mass(
 # %% [markdown]
 # ## Main exercise
 #
-# Define a vector ```weights = [...]``` and a vector ```intercepts = [...]``` of
-# the same length. Each pair of entries ```(weights[i], intercepts[i])``` tags a
+# Define a vector `weights = [...]` and a vector `intercepts = [...]` of
+# the same length. Each pair of entries `(weights[i], intercepts[i])` tags a
 # different model. Use these vectors along with the vector
-# ```flipper_length_range``` to plot several linear models that could possibly
+# `flipper_length_range` to plot several linear models that could possibly
 # fit our data. Use the above helper function to visualize both the models and
 # the real samples.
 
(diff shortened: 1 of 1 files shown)


python_scripts/linear_models_sol_01.py
```diff
@@ -46,9 +46,12 @@ def linear_model_flipper_mass(
...
 #
-# Given a vector of the flipper length, several weights and intercepts to
-# plot several linear model that could fit our data. Use the above
-# helper function to visualize both the model and data.
+# Define a vector ```weights = [...]``` and a vector ```intercepts = [...]``` of
+# the same length. Each pair of entries ```(weights[i], intercepts[i])``` tags a
+# different model. Use these vectors along with the vector
+# ```flipper_length_range``` to plot several linear models that could possibly
+# fit our data. Use the above helper function to visualize both the models and
+# the real samples.
```
glemaitre, line 54: is it normal to have three backticks to define a text as code here?

python_scripts/linear_models_sol_01.py
```diff
@@ -93,10 +96,10 @@ def goodness_fit_measure(true_values, predictions):
     # we compute the error between the true values and the predictions of our
     # model
     errors = np.ravel(true_values) - np.ravel(predictions)
-    # We have several possible strategy to reduce all errors to a single value.
-    # Computing the mean error (sum divided by the number of element) looks
-    # like a good solution. However, we have negative error and therefore, we
-    # need to sum only positive numbers. Therefore, we can either square each
+    # We have several possible strategies to reduce all errors to a single value.
+    # Computing the mean error (sum divided by the number of element) could seem
+    # like a good solution. However, we have negative errors that will misleadingly
```
glemaitre, line 101: "could seem like" seems odd to me :)

ArturoAmorQ: We can alternatively use "[might seem like](https://www.phrasemix.com/phrases/it-might-seem-like-something-but-sentence)", which is a little better. What do you think?

python_scripts/linear_models_sol_01.py
ArturoAmorQ: I had in mind that three back-ticks are good for quoting code in a particular format, though this is probably unnecessary here.

ogrisel: Triple backticks are probably only useful for multi-line code blocks.

python_scripts/linear_models_sol_01.py
ogrisel: +1 for "might seem like" or just "seems like" as you prefer.

glemaitre:
Thanks @ArturoAmorQ

lesteve:
Note to ourselves: each time we change a solution we need to regenerate the corresponding exercise.

lesteve:
I just pushed a commit updating the exercise

PR #431, opened 2021-08-04
Fix wording in linear models exercise for clarity

ArturoAmorQ:
Instructions in linear_models_ex_03.py were unclear. Rephrased to fix it and changed the order of two instructions to improve linearity of exposition.

### DIFF
--- a/python_scripts/linear_models_ex_03.py
+++ b/python_scripts/linear_models_ex_03.py
@@ -29,17 +29,17 @@ target *= 100  # rescale the target in k$
 data.head()
 
 # %% [markdown]
-# Now this is your turn to train a linear regression model on this dataset.
+# Now it is your turn to train a linear regression model on this dataset.
 # You will need to:
 # * create a linear regression model;
 # * execute a cross-validation with 10 folds and use the mean absolute error
-#   (MAE) as metric. Ensure to return the fitted estimators;
-# * compute mean and std of the MAE in thousands of dollars (k$);
-# * show the values of the coefficients for each feature using a boxplot by
-#   inspecting the fitted model returned from the cross-validation. Hint: you
+#   (MAE) as metric. Be sure to *return* the fitted *estimators*;
+# * inspect the fitted model using a box plot to show the distribution of values
+#   for the coefficients returned from the cross-validation. Hint:
 #   use the function
 #   [`df.plot.box()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.box.html)
-#   to plot a box plot.
+#   to create a box plot.
+# * compute mean and std of the MAE in thousands of dollars (k$);
 
 # %%
 # Write your code here.
--- a/python_scripts/linear_models_sol_03.py
+++ b/python_scripts/linear_models_sol_03.py
@@ -28,17 +28,17 @@ target *= 100  # rescale the target in k$
 data.head()
 
 # %% [markdown]
-# Now this is your turn to train a linear regression model on this dataset.
+# Now it is your turn to train a linear regression model on this dataset.
 # You will need to:
 # * create a linear regression model;
 # * execute a cross-validation with 10 folds and use the mean absolute error
-#   (MAE) as metric. Ensure to return the fitted estimators;
-# * compute mean and std of the MAE in thousands of dollars (k$);
-# * show the values of the coefficients for each feature using a boxplot by
-#   inspecting the fitted model returned from the cross-validation. Hint: you
+#   (MAE) as metric. Be sure to *return* the fitted *estimators*;
+# * inspect the fitted model using a box plot to show the distribution of values
+#   for the coefficients returned from the cross-validation. Hint:
 #   use the function
 #   [`df.plot.box()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.box.html)
-#   to plot a box plot.
+#   to create a box plot.
+# * compute mean and std of the MAE in thousands of dollars (k$);
 
 # %%
 # solution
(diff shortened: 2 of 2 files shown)


python_scripts/linear_models_ex_03.py
```diff
@@ -29,17 +29,17 @@
...
-# * show the values of the coefficients for each feature using a boxplot by
-#   inspecting the fitted model returned from the cross-validation. Hint: you
+#   (MAE) as metric. Be sure to *return* the fitted *estimators*;
+# * inspect the fitted model using a box plot to show the distribution of values
+#   for the coefficients returned from the cross-validation. Hint:
 #   use the function
 #   [`df.plot.box()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.box.html)
-#   to plot a box plot.
+#   to create a box plot.
+# * compute mean and std of the MAE in thousands of dollars (k$);
```
lesteve, line 42: Honestly I think we should stop using `;` in bullet points ...

If you want to follow @glemaitre on this one the previous bullet point should have a `;` and the final one a `.`

This is a good occasion for me to quote something I read recently (completely unrelated to the MOOC) that made me laugh:

> Guerre au point-virgule ! Ce parasite, ce timoré, cet affadisseur qui ne marque que l'incertitude, le manque d'audace, le flou de la pensée, et colle aux dents du lecteur comme du caramel mou !

The original quote seems to be from François Cavanna in Mignonne allons voir si la rose ...

ArturoAmorQ: Thanks! Fixed it.

lesteve:
To be honest in an ideal world, this kind of exercise should be split in more or less independent parts like this:
- question 1
- solution 1
- question 2
- solution 2
- question 3
- solution 3

rather than:
- questions 1, 2, 3 in bullet points
- solution for all the questions mixed together

In my opinion this makes it easier to try each part and potentially get back on track if you don't manage to tackle a question. Otherwise this is too hard and basically you try a bit and then give up and go see the full solution.

The fact that no one complained in the forum about exercises makes me think:
- people went to read directly the solution
- maybe bothering about this is not a useful way to spend our time, this is not important enough for how people use the MOOC

lesteve:
Also just curious, why did you decide to move the mean + std question at the end? It feels this is more closely linked to `cross_validate` (at least the solution is more straightforard once you have used `cross_validate` whereas the `coef_` question is a bit more separate)

ArturoAmorQ:
> The fact that no one complained in the forum about exercises makes me think:
> 
>  -   people went to read directly the solution
>   -  maybe bothering about this is not a useful way to spend our time, this is not important enough for how people use the MOOC

My personal experience (and maybe with not enough generalization performance) was that the instructions were so hard to follow that discouraged me from trying to code it and I rather clicked on the solution to understand what was being asked :disappointed: Nobody would accept in the forum that they did so!

> Also just curious, why did you decide to move the mean + std question at the end? It feels this is more closely linked to cross_validate (at least the solution is more straightforard once you have used cross_validate whereas the coef_ question is a bit more separate)

I think in terms of attention spam it is better to keep the motivation for using `return_estimator=True` close to the actual use. As computing mean and std don't require further motivation it can be done at any point, but doing it at the end may feel as encouraging as a "dessert". I agree it is a matter of taste at the end of the day.

> To be honest in an ideal world, this kind of exercise should be split in more or less independent parts 

That was actually my first thought, but I didn't intended to be too polemic on a wording PR :sweat_smile: but then I will address it!

lesteve:
> My personal experience (and maybe with not enough generalization performance) was that the instructions were so hard to follow that discouraged me from trying to code it and I rather clicked on the solution to understand what was being asked disappointed Nobody would accept in the forum that they did so

So that's my feeling as well on this particular exercise ... are you saying this is something that happened for this particular exercise or is it the case for most exercises?

Side-comment: that the exercise with corrections are used during in-person courses where we actually leave some time for people to do the exercise during the session (even then people sometimes have a tendency to stop trying at one point especially towards the end of the session ...). During in-person courses, you can react to people problems and guide them a bit.

lesteve:
> I think in terms of attention spam it is better to keep the motivation for using return_estimator=True close to the actual use. As computing mean and std don't require further motivation it can be done at any point, but doing it at the end may feel as encouraging as a "dessert". I agree it is a matter of taste at the end of the day.

One possibility: 
- a first question without return_estimator and make it more about returning the scores. My feeling is that the primary goal of doing a cross-validation is to look at the scores
- do the second question with return_estimator explaining why you need it in this particular case

ArturoAmorQ:
> are you saying this is something that happened for this particular exercise or is it the case for most exercises?

Yes, I mean this particular exercise.

> During in-person courses, you can react to people problems and guide them a bit.

Sure! This problem arises only for people doing the MOOC at home.

ArturoAmorQ:
> > I think in terms of attention spam it is better to keep the motivation for using return_estimator=True close to the actual use. As computing mean and std don't require further motivation it can be done at any point, but doing it at the end may feel as encouraging as a "dessert". I agree it is a matter of taste at the end of the day.
> 
> One possibility:
> 
>     * a first question without return_estimator and make it more about returning the scores. My feeling is that the primary goal of doing a cross-validation is to look at the scores
> 
>     * do the second question with return_estimator explaining why you need it in this particular case

I put it back to the original order. Splitting the questions already improves retention time (as there is less info per step to retain).

lesteve:
Looks good to me, thanks!

lesteve:
> My personal experience (and maybe with not enough generalization performance) was that the instructions were so hard to follow that discouraged me from trying to code it and I rather clicked on the solution to understand what was being asked :disappointed:

I think if you find the instructions hard to follow or you have no idea how to start the exercise, it is very likely that most of the MOOC audience will feel the same way.

PR #436, opened 2021-08-12
Explain difference between nodes and leaves

ArturoAmorQ:
Using [this notebook](https://gitlab.inria.fr/learninglab/mooc-scikit-learn/mooc-scikit-learn-coordination/-/blob/master/Bilans/detailed_quiz_results_analysis.ipynb) we see that Q2 in quiz M5.02 has a score barely above 0.6. Adding a small paragraph to explain the difference between split nodes and leaves may help clarifying this question.


ogrisel (approved):
LGTM, thanks!

PR #437, opened 2021-08-12
Fix a/an before h

ArturoAmorQ:
Fix "a/an before h" rule (see [here](https://www.gramlee.com/blog/indefinite-articles-a-or-an-before-h-blame-the-confusion-on-the-french/) for example):

> However, words that start with the letter “H” do not follow the rule for consonants. For the letter “H”, the pronunciation dictates the indefinite article:
> 
>     Use “a” before words where you pronounce the letter “H” such as “a hat,” “a house” or “a happy cat.”
>     Use “an” before words where you don’t pronounce the letter “H” such as “an herb,” “an hour,” or “an honorable man.”


GaelVaroquaux (approved):
Looks great to me. Thanks!!!

Issue #439, opened 2021-08-18
Explaining special case of nested-CV in M3


ArturoAmorQ:
The need for a train-validation-test split for parameter tuning is introduced in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_manual.html) at the beginning of M3 as a _warning_, but using CV within the training set and then scoring it in a fixed test set (as in [this figure](https://scikit-learn.org/stable/_images/grid_search_cross_validation.png)) is a practice that was never introduced yet used subsequently. This has led to confusion as in these posts in the forum:

https://mooc-forums.inria.fr/moocsl/t/suggestion-presenting-nested-cross-validation/4418/3

https://mooc-forums.inria.fr/moocsl/t/m7-conclusion-best-workflow/4711

This method should be introduced at the very beginning of the M3. It would be even better if we use a **video** to dynamically show the swapping folds and explain in depth said _warning_. Doing so will also motivate nested-CV as a generalization when the outer loop has more than 1-fold (no longer fixed test set).

PR #440, opened 2021-08-19
Add scoring model on test set

ArturoAmorQ:
In the same line of thought as in #439, after finding the best parameter on the train-validation sets in [Exercise M3.01](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_ex_02.html) and its [solution](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_sol_02.html), we should ask the student to score it on the test set. Else the role of the initial train-test split is unclear.


ogrisel (approved):
LGTM, just a few comments:

PR #454, opened 2021-09-13
Add illustration of the effect of scaling data

### LINKED ISSUE #446: Show illustration of `StandardScaler()` in lecture notebook
Using the [notebook](https://gitlab.inria.fr/learninglab/mooc-scikit-learn/mooc-scikit-learn-coordination/-/blob/master/Bilans/detailed_quiz_results_analysis.ipynb) introduced in #432, we find that Q4 and Q5 from M1.2 have scores under 0.65. Both questions ask for the effect of `StandardScaler()` on data but one is visual (Q4) and the other one (Q5) is not. 

To address this issue while improving the intuition of scaling, I would suggest showing in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/02_numerical_pipeline_scaling.html) an illustration similar to this one:

StandardScaler

Or even adding the piece of code I used for creating that image, so that people can play on their own (the disadvantage would be making this particular notebook too heavy...)

```python
import matplotlib.pyplot as plt
from sklearn.datasets import make_blobs
plt.rcParams["figure.figsize"] = (10,7)
plt.rcParams.update({'font.size': 16}) 

centers = [[0, 2], [3, 0.5]]
X, _ = make_blobs(n_samples=100, n_features=2, centers=centers, 
                 cluster_std=0.5, center_box=(1, 10.0), 
                 shuffle=True, random_state=0)
 
# plot the training points
fig, ax = plt.subplots()
plt.scatter(X[:, 0], X[:, 1])
ax.spines['left'].set_position('zero')
ax.spines['right'].set_color('none')
ax.spines['bottom'].set_position('zero')
ax.spines['top'].set_color('none')
plt.show()
```

```python
from sklearn.preprocessing import StandardScaler

scaler = StandardScaler()
scaler.fit(X)
X_scaled = scaler.transform(X)

fig, ax = plt.subplots()
plt.scatter(X_scaled[:, 0], X_scaled[:, 1])
ax.spines['left'].set_position('zero')
ax.spines['right'].set_color('none')
ax.spines['bottom'].set_position('zero')
ax.spines['top'].set_color('none')
plt.show()
```

What do you think?

ArturoAmorQ:
closes #446 

Adds code to visualize the effect of `StandardScaler` on gaussian blobs created with `sklearn.datasets.make_blobs`

### DIFF
--- a/python_scripts/02_numerical_pipeline_scaling.py
+++ b/python_scripts/02_numerical_pipeline_scaling.py
@@ -192,20 +192,27 @@ data_train_scaled = pd.DataFrame(data_train_scaled,
 data_train_scaled.describe()
 
 # %% [markdown]
-# We can use a jointplot to visualize the histograms and scatterplot of any
-# pair of numerical features at the same time. We can observe that `StandardScaler`
-# does not change the structure of the data itself but the axes
-# get shifted and scaled.
+# Notice that the mean of all the columns is close to 0 and the standard deviation
+# in all cases is close to 1.
+# We can also visualize the effect of `StandardScaler` using a jointplot to show
+# both the histograms of the distributions and a scatterplot of any pair of numerical
+# features at the same time. We can observe that `StandardScaler` does not change
+# the structure of the data itself but the axes get shifted and scaled.
 
 # %%
 import matplotlib.pyplot  as plt
 import seaborn as sns
 
-sns.jointplot(data=data_train[:300], x="age", y="hours-per-week", marginal_kws=dict(bins=15))
-plt.suptitle("Jointplot of 'age' vs 'hours-per-week' \nbefore StandardScaler", y = 1.1)
+# number of points to visualize to have a clearer plot
+num_points_to_plot = 300
 
-sns.jointplot(data=data_train_scaled[:300], x="age", y="hours-per-week", marginal_kws=dict(bins=15))
-_ = plt.suptitle("Jointplot of 'age' vs 'hours-per-week' \nafter StandardScaler", y = 1.1)
+sns.jointplot(data=data_train[:num_points_to_plot], x="age",
+              y="hours-per-week", marginal_kws=dict(bins=15))
+plt.suptitle("Jointplot of 'age' vs 'hours-per-week' \nbefore StandardScaler", y=1.1)
+
+sns.jointplot(data=data_train_scaled[:num_points_to_plot], x="age",
+              y="hours-per-week", marginal_kws=dict(bins=15))
+_ = plt.suptitle("Jointplot of 'age' vs 'hours-per-week' \nafter StandardScaler", y=1.1)
 
 # %% [markdown]
 # We can easily combine sequential operations with a scikit-learn
--- a/python_scripts/02_numerical_pipeline_scaling.py
+++ b/python_scripts/02_numerical_pipeline_scaling.py
@@ -192,7 +192,48 @@ data_train_scaled = pd.DataFrame(data_train_scaled,
 data_train_scaled.describe()
 
 # %% [markdown]
-# We can easily combine these sequential operations with a scikit-learn
+# ```{tip}
+# You can play with the code below to get familiar with `StandardScaler`.
+# Try changing the position of the `centers`.
+# ```
+
+# %%
+import matplotlib.pyplot as plt
+from sklearn.datasets import make_blobs
+
+# we generate a synthetic dataset "X" with gaussian blobs
+centers = [[0, 2], [3, 0.5]]
+X, _ =make_blobs(n_samples=100, n_features=2, centers=centers,
+                 cluster_std=0.5, center_box=(1, 10.0),
+                 shuffle=True, random_state=0)
+
+# we "fit" and "transform" the datapoints
+scaler = StandardScaler()
+scaler.fit(X)
+X_scaled = scaler.transform(X)
+
+# plot the synthetic points
+fig, ax = plt.subplots()
+plt.title("Gaussian blobs synthetic data")
+plt.scatter(X[:, 0], X[:, 1])
+ax.spines['left'].set_position('zero')
+ax.spines['right'].set_color('none')
+ax.spines['bottom'].set_position('zero')
+ax.spines['top'].set_color('none')
+plt.show()
+
+# plot the scaled data points
+fig, ax = plt.subplots()
+plt.title("Scaled blobs using StandardScaler")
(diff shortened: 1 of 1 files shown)


glemaitre:
Side note @ArturoAmorQ Instead of adding "Fixes #xxx" in the title, you can add "closes #xxx" in the description of the summary of the first PR message. It will have the advantage that issue #xxx will be close when we are going to merge.

glemaitre (commented):
If we go with this synthetic example, that would be my comments. However I will propose an alternative

glemaitre:
In the first notebook, we did use `plt.hist`. Instead of generating a new dataset a bit dissociated from the notebook, we could focus on a single feature. It avoids having to explain anything regarding Gaussian blobs and will be more intuitive.

In short, we could plot the histogram of the `"age"` columns before and after scaling. Basically, it will give these figures:
draft

```python
# %%
fig, axs = plt.subplots(ncols=2, sharey=True, figsize=(6, 4))
axs[0].hist(data_train["age"], bins=15)
axs[1].hist(data_train_scaled[:, 0], bins=15)
axs[0].set_xlabel("Age")
axs[1].set_xlabel("Normalized Age")
axs[0].set_ylabel("Frequency")
fig.suptitle("Histogram of the age feature \nbefore and after normalization", y=1.05)
```

(I would not use the code above and instead do 2 separate figures to have more basic matplotlib code.)

The nice part is that we show that the distribution does change. Only the x-axis values are changing. We could add use the transformer to transform the minimum age and maximum age on the x-axis to explain how do we get these new values in the scaled data. We can then easily emphasize that the values becomes negative.

ArturoAmorQ:
The idea of this PR and #446 is to show a figure similar to the one we show in Quiz M1.2 Q4, i.e., using two features to create a plot in the Cartesian plane. The goal is to improve the score for both Q4 (64.7%) and Q5 (40.7%).

I am not sure if people will be able to see the connection between an hist plot and said figures. The fact is that the text explanation has shown not to suffice for passing the concept, so I would assume this connection is not that easy to make.

>  It avoids having to explain anything regarding Gaussian blobs and will be more intuitive.

Using the data-set could be a good idea as long as we plot it in the Cartesian plane, but letting people play with the position of the centers is worth it in my opinion. Maybe we can find a turnaround like creating using the data-set and encouraging people to visit [this example in the documentation](https://scikit-learn.org/stable/auto_examples/preprocessing/plot_all_scaling.html), or just adding a fixed image as mentioned in #446. Thoughts on that?

glemaitre:
> The goal is to improve the score for both Q4 (64.7%) and Q5 (40.7%).

65% is fine. The second question is difficult because you need to think that features can become negative and this aspect was discussed in the course before. Thus, I am fine with adding basically the answer in the notebook with the data at hand.  

> I am not sure if people will be able to see the connection between an hist plot and said figures. The fact is that the text explanation has shown not to suffice for passing the concept, so I would assume this connection is not that easy to make.

Adding the new dataset and figure just allow to answer the quiz without not much thinking but it is completely disconnected from the course. I would really not do that just for the sake of people getting the right answer. Instead, we should make it explicit on the data at hand what we missed to explain at first.

> Using the data-set could be a good idea as long as we plot it in the Cartesian plane

The histogram is useful there. It allows showing that there is no change in the distribution indeed. We can still plot thing in the feature space thought. I could imagine a strip plot together with the marginal distribution on the top

> but letting people play with the position of the centers is worth it in my opinion

IMO, he does not bring any values

> Maybe we can find a turnaround like creating using the data-set and encouraging people to visit this example in the documentation, or just adding a fixed image as mentioned in #446. 

This example is quite complex but it is worth to link to it explaining that `StandardScaler` is only an option and other preprocessors are existing.

ArturoAmorQ:
I committed a new illustration using the adult_census data-set to avoid creating an ad hoc data-set with make_blobs.

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -192,7 +192,23 @@
 data_train_scaled.describe()
 
 # %% [markdown]
-# We can easily combine these sequential operations with a scikit-learn
+# We can use a jointplot to visualize the histograms and scatterplot of any
```
lesteve, line 195: I think before this you can add something to say that in the previous cell the mean of all the columns is close to 0 and the standard deviation of all the columns is close to 1 which is the `StandardScaler`'s purpose

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -192,7 +192,23 @@
...
+# pair of numerical features at the same time. We can observe that `StandardScaler`
+# does not change the structure of the data itself but the axes
+# get shifted and scaled.
+
+# %%
+import matplotlib.pyplot  as plt
+import seaborn as sns
+
+sns.jointplot(data=data_train[:300], x="age", y="hours-per-week", marginal_kws=dict(bins=15))
```
lesteve, line 204: Put a comment before saying where the magical 300 comes from something like this:

```py

# number of points to visualize to have a clearer plot
num_points_to_plot = 300

sns.jointplot(data=data_train[:num_points_to_plot], ...)

```

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -192,7 +192,23 @@
...
+plt.suptitle("Jointplot of 'age' vs 'hours-per-week' \nbefore StandardScaler", y = 1.1)
```
lesteve, line 205: PEP8 no space around equal space for parameter `suptitle(..., y = 1.1)` -> `suptitle(..., y=1.1)`

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -192,7 +192,23 @@
...
+
+sns.jointplot(data=data_train_scaled[:300], x="age", y="hours-per-week", marginal_kws=dict(bins=15))
+_ = plt.suptitle("Jointplot of 'age' vs 'hours-per-week' \nafter StandardScaler", y = 1.1)
```
lesteve, line 208: PEP8 no space around equal space for parameter `suptitle(..., y = 1.1)` -> `suptitle(..., y=1.1)`

lesteve (commented):
A few comments

python_scripts/02_numerical_pipeline_scaling.py
ArturoAmorQ: Done, thanks for the comments!

PR #455, opened 2021-09-13
Clarify the notation "data_test" when no train-test split is involved

### LINKED ISSUE #442: Confusing use of term "test"
In the following notebooks the word "test" is used for denoting the range of possible values in a regression and not a subset of the data coming from a train-test split or cross-validation:

- [trees_regression.py](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_regression.html)
- [trees_ex_02.py](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_ex_02.html) and its [solution](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_sol_02.html)
- [trees_hyperparameters.py](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_hyperparameters.html)
- [ensemble_bagging.py](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_bagging.html)
- [ensemble_gradient_boosting.py](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_gradient_boosting.html)

In the notebook [trees_regression.py](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_regression.html) we implicitly justify the use of the word "test" as something over which we can use the  `predict` method. This might be confusing, especially after addressing Issue #441.

In [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_02.html) we use the term `data_ranges` instead of `data_test` to avoid confusion. For the penguins data-set we use the term `flipper_length_range` [here](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_ex_01.html), [here](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_sol_01.html), [here](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_regression_in_sklearn.html) and [here](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_regression_without_sklearn.html).

ArturoAmorQ:
closes #442 

In some notebooks the word "test" is used for denoting data generated in equally-spaced intervals for the visual evaluation of the
predictions, given that such data was not seen during training. This could be confusing with the notion of "test" as a subset of the data coming from either a train-test split or cross-validation.
I changed  `data_test` to `data_range` whenever the former meaning was intended. The notation was chosen to match with [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_02.html), where we use the term `data_ranges` instead of `data_test`. I also edited the rest of the text accordingly and added a warning when _both_ meanings were intended in the same notebook.

### DIFF
--- a/python_scripts/trees_regression.py
+++ b/python_scripts/trees_regression.py
@@ -41,6 +41,14 @@ data_test = pd.DataFrame(np.arange(data_train[data_columns[0]].min(),
 # It should not be confused with data coming from a train-test split, as it
 # was generated in equally-spaced intervals for the visual evaluation of the
 # predictions.
+#
+# Note that this is methodologically valid here because our objective is to get
+# some intuitive understanding on the shape of the decision function of the
+# learned decision trees.
+#
+# However computing an evaluation metric on such a synthetic test set would
+# be meaningless since the synthetic dataset does not follow the same
+# distribution as the real world data on which the model will be deployed.
 
 # %%
 import matplotlib.pyplot as plt
--- a/python_scripts/trees_regression.py
+++ b/python_scripts/trees_regression.py
@@ -32,9 +32,15 @@ data_train, target_train = penguins[data_columns], penguins[target_column]
 # %%
 import numpy as np
 
-data_range = pd.DataFrame(np.arange(data_train[data_columns[0]].min(),
+data_test = pd.DataFrame(np.arange(data_train[data_columns[0]].min(),
                                    data_train[data_columns[0]].max()),
                          columns=data_columns)
+
+# %% [markdown]
+# Using the term "test" here refers to data that was not used for training.
+# It should not be confused with data coming from a train-test split, as it
+# was generated in equally-spaced intervals for the visual evaluation of the
+# predictions.
 
 # %%
 import matplotlib.pyplot as plt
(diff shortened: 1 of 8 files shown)


glemaitre:
I am not super convinced about this change. Indeed, we are manipulating a real test set, i.e. data that we did not see during `fit`. So this is indeed a real test set. However, it has the particularity that it is a synthetic test set where data are linearly spaced. It might be worth emphasizing why we generate such a test set or at least mentioned that we generate such a test set with these particularities. However, `data_ranges` is not a naming that I would expect. I would expect that it contains the min/max of the dataset.

ArturoAmorQ:
I pushed a couple of commits to revert _data_range_ to _data_test_, explain the intended meaning on the term _test_ and, hopefully, improve general wording about this synthetic datasets.

python_scripts/trees_regression.py
```diff
@@ -36,6 +36,12 @@
                                    data_train[data_columns[0]].max()),
                          columns=data_columns)
 
+# %% [markdown]
+# Using the term "test" here refers to data that was not used for training.
+# It should not be confused with data coming from a train-test split, as it
+# was generated in equally-spaced intervals for the visual evaluation of the
+# predictions.
+
```
ogrisel, line 44: Maybe we can repeat the note that I suggested above here. Or just put it here since this section will appear first in the chronological order of the MOOC. As you prefer.

ogrisel (approved):
LGTM, just a small suggestion:

python_scripts/trees_regression.py
ArturoAmorQ: I added the suggestion here and just added a recall of this message in the bagging notebook

glemaitre (approved):
LGTM

PR #459, opened 2021-09-15
Make GBDT overfitting example more illustrative

ArturoAmorQ:
Current example in Exercise M6.04 does not clearly show an increasing gap between training and testing validation curves. 
This fixes it while taking care of keeping short running times.

### DIFF
--- a/python_scripts/ensemble_sol_04.py
+++ b/python_scripts/ensemble_sol_04.py
@@ -17,7 +17,7 @@ from sklearn.model_selection import train_test_split
 data, target = fetch_california_housing(return_X_y=True, as_frame=True)
 target *= 100  # rescale the target in k$
 data_train, data_test, target_train, target_test = train_test_split(
-    data, target, random_state=0, test_size=0.5)
+    data, target, random_state=0, test_size=5000)
 
 # %% [markdown]
 # ```{note}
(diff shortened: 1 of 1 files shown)


python_scripts/ensemble_sol_04.py
```diff
@@ -26,44 +26,57 @@
...
+# Try lowering the number of cross-validation folds to 3 to speed up the evaluation.
+
+# %%
+# solution
 from sklearn.model_selection import validation_curve
 
-gbdt = GradientBoostingRegressor()
-param_range = np.unique(np.logspace(0, 1.8, num=30).astype(int))
-train_scores, test_scores = validation_curve(
+param_range = [1, 2, 5, 10, 20, 50, 100, 200]
```
glemaitre, line 53: Looking at the output, we could maybe remove the last values (200). With 100 iterations, we should have already a good gap.

glemaitre (approved):
Apart of removing the last value, LGTM

python_scripts/ensemble_sol_04.py
ArturoAmorQ: Done, thanks!

lesteve:
I pushed some minor tweaks. This one is good to merge when CI is green

lesteve:
The figure looks like this with this change:
image

PR #460, opened 2021-09-15
Add info about DummyClassifier strategy='stratified'

### LINKED ISSUE #450: Add information about `DummyClassifier` strategy `stratified`
We could add a piece of code in [Solution for Exercise M7.01](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_sol_02.html) to explain and show how the `most_frequent` strategy is deterministic given a cross-validation split, whereas the `stratified` strategy is randomized and will lead to different results even for the same cv split.

Giving a bit more information on `stratified` is asked in [this forum post](https://mooc-forums.inria.fr/moocsl/t/density-argument-in-plt-hist/4191).

ArturoAmorQ:
closes #450 

Added text to expand explanation on `DummyClassifier` `strategy='stratified'`. 
Added loop to show the effect of the random prediction made with this strategy and compare it with the deterministic nature of `strategy='most_frequent'`.


python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,37 @@
...
-# proportion of this class (~75% of the samples). However, by using the
-# `stratified` strategy, wrong predictions will be made even for the most
+# proportion of this class (~75% of the samples). However, the `stratified`
+# strategy will randomly generate predictions by respecting the training
+# set’s class distribution, resulting in wrong predictions even for the most
 # frequent class, hence we obtain a lower accuracy.
+#
+# Notice that the `most_frequent` strategy is deterministic for a given
+# cross-validation split, whereas the `stratified` strategy will lead
+# to different results even for the same cv split due to the extra randomization.
```
glemaitre, line 145: It might be worth to add "... if one does not set the parameter `random_state`."

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,37 @@
...
+
+# %% tags=["solution"]
```
glemaitre, line 147: I am not sure that we need to add an example. We can link to the documentation of the `random_state` parameter that documents the random selection using `"uniform"` and `"stratify"`. I don't think that the `DummyClassifier` being non-deterministic is actually one of the messages to convey in the lecture. A note bringing a bit more insight would suffice, IMO.

ogrisel: I agree. I think we can keep things simple here. Maybe we could even not mention `strategy="stratify"` at all and simplify the exercise to remove the last question.

In my opinion, what we want to convey with this exercise is that it's useful to always compare your classifier performance to `DummyClassifier(strategy="most_frequent")` as a baseline classifier with a prediction function that totally ignores `X_train` and `X_test` yet can reach seemingly high accuracy scores on imbalanced problems.

The other strategies are only useful for more advanced users who might be curious about other ways to define baselines but they are often useless for practitioners compared to `strategy="most_frequent"` unless one is interested in theoretical considerations that are beyond the target audience of the MOOC.

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,37 @@
 # We see that using `strategy="stratified"`, the results are much worse than
 # with the `most_frequent` strategy. Since the classes are imbalanced,
 # predicting the most frequent involves that we will be right for the
-# proportion of this class (~75% of the samples). However, by using the
-# `stratified` strategy, wrong predictions will be made even for the most
+# proportion of this class (~75% of the samples). However, the `stratified`
+# strategy will randomly generate predictions by respecting the training
+# set’s class distribution, resulting in wrong predictions even for the most
 # frequent class, hence we obtain a lower accuracy.
```
ogrisel, line 141:
```suggestion
# proportion of this class (~75% of the samples). However, the `stratified`
# strategy will randomly generate predictions by respecting the training
# set’s class distribution, resulting in some wrong predictions even for
# the most frequent class, hence we obtain a lower accuracy.
```

ogrisel (commented):
I think this PR still has some value but I would rather like to simplify it by removing the end "Notice that the `most_frequent` strategy is deterministic for a given..." and instead linking to https://scikit-learn.org/stable/modules/generated/sklearn.dummy.DummyClassifier.html in the exercise instructions:

```
# Change the strategy of the dummy classifier to `stratified`, compute the
# results and plot the distribution together with the other results. Explain
# why the results get worse.
```

For instance by adding:

```
Please refer to the scikit-learn documentation on [DummyClassifier](https://scikit-learn.org/stable/modules/generated/sklearn.dummy.DummyClassifier.html) for more details on the meaning of the `strategy="stratified"` parameter.
```

ArturoAmorQ:
> I would rather like to simplify it by removing the end "Notice that the most_frequent strategy is deterministic for a given..."

Does that mean removing from there on, or is it worth keeping the final plot?

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,38 @@
...
+        result_most_frequent["test_score"], name="Dummy 'most_frequent' score"
+    )
+    scores_stratified = pd.Series(
+        result_stratified["test_score"], name="Dummy 'stratify' score"
+    )
+
+    final_scores = pd.concat([scores_stratified, scores_most_frequent], axis=1)
+
+    final_scores.plot.hist(bins=50, density=True, edgecolor="black")
+    plt.legend(bbox_to_anchor=(1.05, 0.8), loc="upper left")
```
glemaitre, line 170: It seems that we can simplify by letting the legend in the plot in top center.

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,38 @@
...
+dummy_most_frequent = DummyClassifier(strategy="most_frequent")
+dummy_stratified = DummyClassifier(strategy="stratified")
+n_runs = 3
+
+for run_idx in range(n_runs):
+
+    result_most_frequent = cross_validate(
+        dummy_most_frequent, data, target, cv=cv, n_jobs=2
+    )
+
```
glemaitre, line 157:
```suggestion
```

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,38 @@
...
+# for more details on the meaning of the `strategy="stratified"` parameter.
+
+# %% tags=["solution"]
+dummy_most_frequent = DummyClassifier(strategy="most_frequent")
+dummy_stratified = DummyClassifier(strategy="stratified")
+n_runs = 3
+
+for run_idx in range(n_runs):
+
+    result_most_frequent = cross_validate(
```
glemaitre, line 154: Since we already used `cross_val_score` in the previous notebook and we are only interested in the test score, we can simplify this example by using it.

glemaitre:
```python
dummy_models = {
    "Dummy 'most_frequent'": DummyClassifier(strategy="most_frequent"),
    "Dummy 'stratified'": DummyClassifier(strategy="stratified"),
}
n_runs = 3

for run_idx in range(n_runs):
    final_scores = pd.DataFrame({
        f"{name} score": cross_val_score(model, data, target, cv=cv, n_jobs=2)
        for name, model in dummy_models.items()
    })

    # ... make the plotting
```

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,31 @@
...
+# ](https://scikit-learn.org/stable/modules/generated/sklearn.dummy.DummyClassifier.html)
+# for more details on the meaning of the `strategy="stratified"` parameter.
+
+# %% tags=["solution"]
+from sklearn.model_selection import cross_val_score
+
+dummy_models = {
+    "Dummy 'most_frequent'": DummyClassifier(strategy="most_frequent"),
+    "Dummy 'stratified'": DummyClassifier(strategy="stratified"),
+    }
```
glemaitre, line 153:
```suggestion
}
```

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,31 @@
...
+n_runs = 3
+
+for run_idx in range(n_runs):
+    final_scores = pd.DataFrame({
+        f"{name} score": cross_val_score(model, data, target, cv=cv, n_jobs=2)
+        for name, model in dummy_models.items()
+        })
```
glemaitre, line 160:
```suggestion
    })
```

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,31 @@
 # We see that using `strategy="stratified"`, the results are much worse than
 # with the `most_frequent` strategy. Since the classes are imbalanced,
 # predicting the most frequent involves that we will be right for the
-# proportion of this class (~75% of the samples). However, by using the
-# `stratified` strategy, wrong predictions will be made even for the most
-# frequent class, hence we obtain a lower accuracy.
+# proportion of this class (~75% of the samples). However, the `stratified`
```
glemaitre, line 138:
```suggestion
# proportion of this class (~75% of the samples). However, the `"stratified"`
```

python_scripts/cross_validation_sol_02.py
```diff
@@ -135,6 +135,31 @@
...
+# strategy will randomly generate predictions by respecting the training
+# set’s class distribution, resulting in some wrong predictions even for
```
glemaitre, line 140:
```suggestion
# set's class distribution, resulting in some wrong predictions even for
```

glemaitre (approved):
Otherwise LGTM.

ogrisel (approved):

ogrisel:
Let's wait for the build to complete to check the HTML rendering and then merge.

PR #462, opened 2021-09-16
Add explanation of dates relation to age

ArturoAmorQ:
Wrap-up M1 Q4 has a score of 61.4% This shows that dates can easily be confused with an ordinal category. 
This PR adds an explanation on how date and age are related.

### DIFF
--- a/python_scripts/03_categorical_pipeline_column_transformer.py
+++ b/python_scripts/03_categorical_pipeline_column_transformer.py
@@ -55,10 +55,10 @@ categorical_columns = categorical_columns_selector(data)
 # ```{caution}
 # Here, we know that `object` data type is used to represent strings and thus
 # categorical features. Be aware that this is not always the case. Sometimes
-# `object` data type could contain other type of information, such as dates that
-# were not parsed and can then relate to age (a quantity). In a more general
-# scenario you should manually introspect the content of your dataframe not to
-# wrongly use `make_column_selector`.
+# `object` data type could contain other types of information, such as dates that
+# were not properly formated (strings) and yet can relate to age (a quantity).
+# In a more general scenario you should manually introspect the content of your
+# dataframe not to wrongly use `make_column_selector`.
 # ```
 
 # %% [markdown]
--- a/python_scripts/03_categorical_pipeline_column_transformer.py
+++ b/python_scripts/03_categorical_pipeline_column_transformer.py
@@ -56,7 +56,9 @@ categorical_columns = categorical_columns_selector(data)
 # Here, we know that `object` data type is used to represent strings and thus
 # categorical features. Be aware that this is not always the case. Sometimes
 # `object` data type could contain other types of information, such as dates that
-# were not properly formated (strings) and yet can relate to age (a quantity).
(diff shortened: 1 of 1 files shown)


python_scripts/03_categorical_pipeline_column_transformer.py
```diff
@@ -55,9 +55,10 @@
 # ```{caution}
 # Here, we know that `object` data type is used to represent strings and thus
 # categorical features. Be aware that this is not always the case. Sometimes
-# `object` data type could contain other type of information (e.g. dates that
-# were not parsed) and you should manually introspect the content of your
-# dataframe to not wrongly use `make_column_selector`.
+# `object` data type could contain other type of information, such as dates that
+# were not parsed and can then relate to age (a quantity). In a more general
```
glemaitre, line 59: A couple of thoughts about this change. I don't think that this is enough to remove the ambiguity in the quiz. I would be inclined to get a small paragraph to add more information.

The quiz refers to a "year" and not a date indeed. The year can already be encoded as a numerical value and thus a quantity.

A date that is not parsed properly will be a string, with a format such as "dd-mm-yyyy". It is indeed what we are referring to in the text here. So this is not a quantity. If you use an ordinal encoder, you will transform it to a quantity but it will most probably go sideways. Indeed, the ordering might not be respected and then the encoded value will not be linked to age.

So in the current paragraph, we could make things more explicit regarding the date format (what I try to explain in the previous paragraph). In addition, we can add a new note regarding quantity as year/day/month and why sometimes some numerical values could be considered either a category or a numerical value.

lesteve:
Personally I think the quiz poor scores are more due to the fact that all the columns in the answers are numbers in the dataframe. Using categories that are strings would make the question easier (there are plenty of them).

Since we removed `education-num` we don't really talk about categories already encoded in the dataset so I think this question is a bit harsh.

That does not really solve how to make the warning in this notebook clearer though ...

ArturoAmorQ:
I commited a bit of wording with the hope that this will help improve the clarity.
About Wrap-up M1 Q4 we kind of agreed to omit the option related to dates.

lesteve:
> About Wrap-up M1 Q4 we kind of agreed to omit the option related to dates.

What I am suggesting on top of this is also to use categorical columns that are represented as strings.

python_scripts/03_categorical_pipeline_column_transformer.py
```diff
@@ -55,9 +55,10 @@
 # ```{caution}
 # Here, we know that `object` data type is used to represent strings and thus
 # categorical features. Be aware that this is not always the case. Sometimes
-# `object` data type could contain other type of information (e.g. dates that
-# were not parsed) and you should manually introspect the content of your
-# dataframe to not wrongly use `make_column_selector`.
+# `object` data type could contain other types of information, such as dates that
+# were not properly formated (strings) and yet can relate to age (a quantity).
```
ogrisel, line 59:
```suggestion
# were not properly formatted (strings) and yet relate to a quantity of
# elapsed  time.
#
```

PR #463, opened 2021-09-17
ENH enhanced nested cross-validation in the parameter search

glemaitre:
Introduce better the nested cross-validation. We first introduce cross-validation in the context of hyperparameter tuning. We still keep the nested cross-validation in the last module to focus on its influence on generalization performance.
In this way, the change is minimal. We could decide later if we want to merge the two sections (IMO, it can of work this way for the moment).


python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+# `cross_validate` function to evaluate our model.
 
 # %%
 from sklearn.model_selection import cross_validate
+
+cv_results = cross_validate(model, data, target, cv=5)
+cv_results = pd.DataFrame(cv_results)
+cv_results
+
+# %%
```
ArturoAmorQ, line 88: I guess this should be a [markdown] cell

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
 # %% [markdown]
-# Running the above cross-validation will give us an estimate of the
-# testing score.
+# One important bottleneck here concerns the generalization performance.
+# Indeed, the mean and standard deviation of the generalization score provided
+# by the cross-validation in the grid-search is potentially not a good estimate
+# of the generalization performance of the model refitted with the best
+# parameters (i.e. the model at hand when using `model_grid_search.predict`).
+#
+# The reason is that the model was refitted on the full dataset given during
```
ArturoAmorQ, line 147: Maybe a matter of taste, but I think the argument in the cross_validation_nested.py notebook is clearer:

>  we should not forget that we used this score to pick-up the best model. It means that we used knowledge from the test set (i.e. test score) to decide our model’s training parameter.

ArturoAmorQ:
LGTM! But now that you are editing this file, would you mind changing _trustful_ to _trustworthy_ in the cross_validation_nested.py file, please?

(see [here](https://diffsense.com/diff/trustful/trustworthy) for the use of both adjectives)

ogrisel (commented):
Here is a first pass of review for the first part:

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+    "rank_test_score"
+]]
+
+# %%
+model_grid_search.best_params_
 
 # %% [markdown]
-# Running the above cross-validation will give us an estimate of the
-# testing score.
+# One important bottleneck here concerns the generalization performance.
```
ogrisel, line 141:
```suggestion
# One important caveat here concerns the evaluation of the generalization performance.
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+# Indeed, the mean and standard deviation of the generalization score provided
```
ogrisel, line 142:
```suggestion
# Indeed, the mean and standard deviation of the scores computed
```

python_scripts/parameter_tuning_nested.py
ogrisel: I think both points are interesting (using the best mean internal CV score to select the model to refit and the fact that we refit on the full data after selecting the best model).

However I think it confusing to mention "the test set" in the context of a cross-validation since there are several inner test sets.

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+# by the cross-validation in the grid-search is potentially not a good estimate
+# of the generalization performance of the model refitted with the best
+# parameters (i.e. the model at hand when using `model_grid_search.predict`).
+#
+# The reason is that the model was refitted on the full dataset given during
+# `fit`. Therefore, this refitted model was trained with more data than the
+# different models trained during the cross-validation in the grid-search.
+#
+# Therefore, one must keep a test set to evaluate the refitted model. We
```
ogrisel, line 151: Maybe make it even more explicit:

```suggestion
# Therefore, one must keep an external, held-out test set for the final evaluation the refitted model. We
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
-# The hyperparameters on each fold are potentially different since we nested
-# the grid-search in the cross-validation. Thus, checking the variation of the
-# hyperparameters across folds should also be analyzed.
+# In the code above, the selection of the best hyperparameters was done only on
+# the train set. Then, we evaluated the generalization performance of our
+# tuned model on the left out test set.
+#
+# However, this evaluation only provides us a single point estimate of the
+# generalization performance. As recall at the beginning of this notebook,
+# it is beneficial to have an uncertainty measure of the generalization
```
ogrisel, line 172:
```suggestion
# it is beneficial to have a rough idea of the uncertainty of our estimate of the generalization
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+print(
+    "Generalization score with hyperparameters tuning:\n"
+    f"{cv_results['test_score'].mean():.3f} +/- {cv_results['test_score'].std():.3f}"
+)
 
 # %% [markdown]
-# Obtaining models with unstable hyperparameters would be an issue in practice.
-# Indeed, it would become difficult to set them.
+# In this case, we obtain a distribution of scores and therefore, we can
+# apprehend the variability of the generalization performance.
```
ogrisel, line 198:
```suggestion
# apprehend the variability of our estimate of the generalization performance.
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+#
+# In addition, passing the parameter `return_estimator=True`, we can check the
+# value of the best hyperparameters for each fold of the outer
```
ogrisel, line 201:
```suggestion
# value of the best hyperparameters obtained for each fold of the outer
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +66,155 @@
...
+        f"{estimator_in_fold.best_params_}"
+    )
 
 # %% [markdown]
+# Such an analysis allows to see if only a single set of hyperparameters is
+# sufficient to obtain a good generalization performance or if there is some
+# variation in the hyperparameters used.
+#
+# Obtaining models with unstable hyperparameters would be an issue in practice.
+# Indeed, it would become difficult to set them.
```
ogrisel, line 217: Maybe the following is more operation:

```suggestion
# It is interesting to see whether the hyper-parameter tuning procedure always select
# similar values for the hyper-parameters. If its the case, then all is fine. It means that
# we can deploy a model fit with those hyper-parameters and expect that it will have
# an actual predictive performance close to what we measured in the outer
# cross-validation.
#
# But it is also possible that some hyper-parameters do not matter at all, and as a
# result in different tuning sessions give different results. In this case,
# any value will do. This can typically be confirmed by doing a parallel coordinate plot
# of the results of a large hyper-parameter search as seen in the exercises.
#
# From a deployment, one could also chose to deploy all the models found by the outer
# cross-validation loop and make them vote to get the final predictions. However this
# can cause operational problems because it uses more memory and makes
# computing prediction slower resulting in a higher computational resource usage
# per prediction. 
```

ogrisel (commented):
More comments/suggestions, this time for the part of nested CV and hyper-parameter tuning.

ArturoAmorQ (approved):

glemaitre (commented):
Thanks for taking care of the corrections. The narration is much better.
LGTM for me but maybe @ogrisel wants to have a final look.

glemaitre:
I miss the legend for the legend in the figure to make them entirely self-explanatory.

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +64,187 @@
...
+# them from the test set in blue.
+# ```
+#
+# However, this evaluation only provides us a single point estimate of the
+# generalization performance. As recall at the beginning of this notebook,
+# it is beneficial to have a rough idea of the uncertainty of our estimated
+# generalization performance. Therefore, we should instead use an additional
+# cross-validation for this evaluation.
+#
+# This pattern is called *nested cross-validation*. We use an inner
```
glemaitre, line 186:
```suggestion
# This pattern is called **nested cross-validation**. We use an inner
```

glemaitre (commented):
Still LGTM otherwise.

ogrisel:
I am doing a final review of the rendered jupyter book but I just wanted to say that the new diagrams are very nice.

python_scripts/cross_validation_nested.py
```diff
@@ -34,27 +35,28 @@
...
+search = GridSearchCV(
+    estimator=model_to_tune, param_grid=param_grid, n_jobs=2
+)
 search.fit(data, target)
 
 # %% [markdown]
-# We recall that `GridSearchCV` will train a model with some specific parameter
-# on a training set and evaluate it on testing. However, this evaluation is
-# done via cross-validation using the `cv` parameter. This procedure is
+# We recall that `GridSearchCV` will train a model with some specific parameters
```
ogrisel, line 44:
```suggestion
# We recall that `GridSearchCV` trains a model with some specific parameters
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -1,46 +1,43 @@
...
-
-set_config(display="diagram")
+# Thus, we will first load the dataset and create the predictive model that
+# we want to optimize and later on, evaluate.
+#
+# ## Loading the dataset
+#
+# As in the previous notebook, we will load the Adult census dataset. The
+# loaded dataframe will be split to get the data and the target into two
+# separated variables. In addition, we will drop the column `"education-num"`
```
ogrisel, line 19:
```suggestion
# separated variables. In addition, we drop the column `"education-num"`
```

python_scripts/parameter_tuning_nested.py
```diff
@@ -69,56 +64,189 @@
...
+# we can compute the mean and standard deviation of the generalization score.
+
+# %%
+print(
+    "Generalization score without hyperparameters tuning:\n"
+    f"{cv_results['test_score'].mean():.3f} +/- {cv_results['test_score'].std():.3f}"
+)
+
+# %% [markdown]
+# Now, we will present how to evaluate the model with hyperparameter tuning,
```
ogrisel, line 96:
```suggestion
# We now present how to evaluate the model with hyperparameter tuning,
```

ogrisel (commented):
Style: remove most of the use of the future tense when not necessary and use the present tense instead to lighten the style.

ogrisel:
Merged!

PR #464, opened 2021-09-24
Scoring model as a good practice

### LINKED ISSUE #444: Scoring model as a good practice
Though in almost all of the notebooks we score the presented model somehow, there are some exceptions:

- [logistic_regression.py](https://inria.github.io/scikit-learn-mooc/python_scripts/logistic_regression.html) (not really important as we have an intuition _by eye_ but would be a _nice-to-have_, or least we can add a _Warning message_, I think)
- [linear_models_ex_05.py](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_ex_05.html) and [its solution](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_sol_05.html)
- [trees_hyperparameters.py](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_hyperparameters.html)

On the same line of thought, [ensemble_hyperparameters.py](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_hyperparameters.html) computes cv scores for parameter tuning but it doesn't pass the best parameters to a final test-set scoring.

Plotting a validation curve using the train set but no further scoring on the test-set (as in [Solution for Exercise M6.03](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_03.html) and [Solution for Exercise M6.04](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_04.html)) is discussed in [this forum post](https://mooc-forums.inria.fr/moocsl/t/exm6-03-validation-curve-is-done-with-the-data-train-instead-of-the-whole-data/4309). In this case we can either use the whole data-set or add a _Warning message_.

We also don't score most of the notebooks that use the full data-set for modeling (see #441), but that is perfectly fine.

ArturoAmorQ:
Closes #444

Though in almost all of the notebooks we score the presented model somehow, there are some exceptions:

- [logistic_regression.py](https://inria.github.io/scikit-learn-mooc/python_scripts/logistic_regression.html) (not really important as we have an intuition _by eye_ but would be a _nice-to-have_, or least we can add a _Warning message_, I think)
- [linear_models_ex_05.py](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_ex_05.html) and [its solution](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_sol_05.html)
- [trees_hyperparameters.py](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_hyperparameters.html)

On the same line of thought, [ensemble_hyperparameters.py](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_hyperparameters.html) computes cv scores for parameter tuning but it doesn't pass the best parameters to a final test-set scoring.

Plotting a validation curve using the train set but no further scoring on the test-set (as in [Solution for Exercise M6.03](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_03.html) and [Solution for Exercise M6.04](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_04.html)) is discussed in [this forum post](https://mooc-forums.inria.fr/moocsl/t/exm6-03-validation-curve-is-done-with-the-data-train-instead-of-the-whole-data/4309). The former is not addressed in this PR as the AdaBoost notebook will be fixed elsewhere, the latter was solved by computing asking the student to score the model with `n_iter_no_change=5`.


python_scripts/ensemble_hyperparameters.py
```diff
@@ -62,8 +62,15 @@
 
 # %% [markdown]
 # We can observe that in our grid-search, the largest `max_depth` together
-# with the largest `n_estimators` led to the best generalization performance.
-#
+# with the largest `n_estimators` led to the best performance in the validation
+# set. Now we will estimate the generalization performance of the best model
```
glemaitre, line 66: Since we have cross-validation, then we have several validations sets. So we should use the plural when referring to the set and probably mentioned that the ranking is based on the average.

```
led in average to the best performance on the validation sets
```

python_scripts/ensemble_hyperparameters.py
```diff
@@ -62,8 +62,15 @@
...
+# using the test set.
```
glemaitre, line 67: We might make it explicit that we will use the test set that has not been used and that with the model fitted on the full training set with the best hyperparameters found earlier.

python_scripts/ensemble_hyperparameters.py
```diff
@@ -121,3 +128,10 @@
 # Here, we tune the `n_estimators` but be aware that using early-stopping as
 # in the previous exercise will be better.
 # ```
+
+# %% [markdown]
+# Now we estimate the generalization performance of the best model
+# using the test set.
```
glemaitre, line 134:
```suggestion
# using the test set.

```

python_scripts/linear_models_sol_05.py
```diff
@@ -108,6 +108,9 @@ def plot_decision_function(fitted_classifier, range_features, ax=None):
         hue=target_column, palette=["tab:red", "tab:blue"])
     plot_decision_function(logistic_regression, range_features, ax=ax)
     plt.title(f"C: {C}")
+    plt.show()
```
glemaitre, line 111: We can use a cell for the plot and a cell to compute the score
```suggestion
    
    # %%
```

python_scripts/linear_models_sol_05.py
```diff
@@ -108,6 +108,9 @@ def plot_decision_function(fitted_classifier, range_features, ax=None):
...
+    accuracy = logistic_regression.score(data_test, target_test)
+    print(f"Accuracy on test set: {accuracy:.3f}")
```
glemaitre, line 113:
```suggestion
    print(f"Accuracy on the test set: {accuracy:.3f}")
```

python_scripts/logistic_regression.py
```diff
@@ -124,7 +124,10 @@ def plot_decision_function(fitted_classifier, range_features, ax=None):
 ax = sns.scatterplot(
     data=penguins_test, x=culmen_columns[0], y=culmen_columns[1],
     hue=target_column, palette=["tab:red", "tab:blue"])
-_ = plot_decision_function(logistic_regression, range_features, ax=ax)
+plot_decision_function(logistic_regression, range_features, ax=ax)
+plt.show()
+accuracy = logistic_regression.score(data_test, target_test)
```
glemaitre, line 129: We can move the computation of the accuracy score in the cell above just after the `fit` and let the plotting to be as-is in a single cell.

python_scripts/trees_hyperparameters.py
```diff
@@ -115,12 +115,20 @@ def plot_regression(model, X, y, ax=None):
 plot_classification(tree_clf, data_clf[data_clf_columns],
                     data_clf[target_clf_column])
 plt.legend(bbox_to_anchor=(1.05, 1), loc='upper left')
-_ = plt.title(f"Shallow classification tree with max-depth of {max_depth}")
+plt.title(f"Shallow classification tree with max-depth of {max_depth}")
+plt.show()
+accuracy = tree_clf.score(data_clf[data_clf_columns],
```
glemaitre, line 120: Here I would integrate the scoring in the plot directly.
We should check if it is better to put it as a title or rather in the legend of the plot.

glemaitre: I would have the same comment for the other changes in this file

glemaitre (commented):
Otherwise looks good

python_scripts/linear_models_sol_05.py
ArturoAmorQ: The score is computed inside a for loop that starts in line 101. The goal was to score different values for `C` on the fly.

ArturoAmorQ: Nevermind, I will add a legend to each plot containing the score as suggested below. Thanks!

python_scripts/ensemble_hyperparameters.py
```diff
@@ -62,8 +62,17 @@
 
 # %% [markdown]
 # We can observe that in our grid-search, the largest `max_depth` together
-# with the largest `n_estimators` led to the best generalization performance.
-#
+# with the largest `n_estimators` led in average to the best performance
+# on the validation sets. Now we will estimate the generalization performance
+# of the best model by refitting it with the full training set and using the test
```
glemaitre, line 67: I assume this is the discussion that we had at launch IRL. Here, we need a rephrasing because it seems that the refitting is done in `score`. However, this is done only in the previous call to `fit`.

python_scripts/trees_hyperparameters.py
```diff
@@ -114,13 +114,25 @@ def plot_regression(model, X, y, ax=None):
...
-                data_reg[target_reg_column])
-_ = plt.title(f"Shallow regression tree with max-depth of {max_depth}")
+plot_regression(tree_reg, data_reg[data_reg_columns], data_reg[target_reg_column])
+accuracy = tree_reg.score(data_reg[data_reg_columns], data_reg[target_reg_column])
+
+plt.title(
+    f"Shallow regression tree with max-depth of {max_depth}"
+    f"\n R$^2$ of the fit: {accuracy:.2f}"
+)
+plt.show()
```
glemaitre, line 135: Do we need `plt.show()` here (and later)?

glemaitre (approved):
Otherwise LGTM

python_scripts/ensemble_hyperparameters.py
ArturoAmorQ: You are right, I forgot to edit this notebook after our discussion, Thanks!

ArturoAmorQ:
Comments addressed, thanks!

python_scripts/ensemble_hyperparameters.py
```diff
@@ -121,3 +127,11 @@
...
 # in the previous exercise will be better.
 # ```
+
+# %% [markdown]
+# Now we estimate the generalization performance of the best model
+# using the test set.
+
+# %%
+score = -grid_search.score(data_test, target_test)
+print(f"On average, our GBDT regressor makes an error of {score:.2f} k$")
```
ogrisel, line 137: Here again it would be nice to add a comment to quickly compare this value to the CV values observed during parameter tuning (they should match approximately but it can be even slightly better because the final model has been retrained on more that the inner CV models of the grid search procedure).

ogrisel (commented):
There are conflicts to solve and in general I think we should add an analysis of the new test scores.

python_scripts/ensemble_sol_04.py
```diff
@@ -103,4 +102,14 @@
...
+
+# %% [markdown]
+# Estimate the generalization performance of this model using the test set
+# and `sklearn.metrics.mean_absolute_error`.
+
+# %%
+# solution
+from sklearn.metrics import mean_absolute_error
+score = mean_absolute_error(target_test, gbdt.predict(data_test))
+print(f"On average, our GBDT regressor makes an error of {score:.2f} k$")
```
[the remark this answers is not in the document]
ArturoAmorQ: Thanks, done!

python_scripts/ensemble_hyperparameters.py
ArturoAmorQ: Thanks, done!

python_scripts/ensemble_hyperparameters.py
```diff
@@ -61,22 +61,29 @@
 cv_results[columns].sort_values(by="rank_test_score")
 
 # %% [markdown]
-# We can observe that in our grid-search, the largest `max_depth` together
-# with the largest `n_estimators` led to the best generalization performance.
-#
+# We can observe that in our grid-search, the largest `max_depth` together with
+# the largest `n_estimators` led in average to the best performance on the
```
ogrisel, line 65:
```suggestion
# the largest `n_estimators` led, on average, to the best performance on the
```

python_scripts/ensemble_hyperparameters.py
```diff
@@ -61,22 +61,29 @@
...
+# validation sets. Now we will estimate the generalization performance of the
+# best model by refitting it with the full training set and using the test set
+# for scoring on unseen data. This is done by default when calling the `.fit`
+# method.
+
+# %%
+score = -grid_search.score(data_test, target_test)
+print(f"On average, our random forest regressor makes an error of {score:.2f} k$")
```
ogrisel, line 73:
```suggestion
error = -grid_search.score(data_test, target_test)
print(f"On average, our random forest regressor makes an error of {error:.2f} k$")
```

python_scripts/ensemble_hyperparameters.py
ogrisel:
```suggestion
print(f"On average, our GBDT regressor makes an error of {error:.2f} k$")
```

python_scripts/ensemble_hyperparameters.py
```diff
@@ -121,3 +127,17 @@
 # Here, we tune the `n_estimators` but be aware that using early-stopping as
 # in the previous exercise will be better.
 # ```
+
+# %% [markdown]
+# Now we estimate the generalization performance of the best model
+# using the test set.
+
+# %%
+score = -grid_search.score(data_test, target_test)
```
ogrisel, line 136:
```suggestion
error = -grid_search.score(data_test, target_test)
```

python_scripts/ensemble_sol_04.py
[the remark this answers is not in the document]
ogrisel:
```suggestion
print(f"On average, our GBDT regressor makes an error of {error:.2f} k$")
```

python_scripts/ensemble_sol_04.py
```diff
@@ -103,4 +102,29 @@
...
+# Estimate the generalization performance of this model again using
+# the `sklearn.metrics.mean_absolute_error` metric but this time using
+# the test set that we held out at the beginning of the notebook.
+# Compare the resulting value with the values observed in the validation
+# curve.
+
+# %%
+# solution
+from sklearn.metrics import mean_absolute_error
+score = mean_absolute_error(target_test, gbdt.predict(data_test))
```
ogrisel, line 117:
```suggestion
error = mean_absolute_error(target_test, gbdt.predict(data_test))
```

python_scripts/trees_hyperparameters.py
```diff
@@ -114,13 +114,23 @@ def plot_regression(model, X, y, ax=None):
...
                     data_clf[target_clf_column])
+accuracy = tree_clf.score(data_clf[data_clf_columns],
+                          data_clf[target_clf_column])
+
 plt.legend(bbox_to_anchor=(1.05, 1), loc='upper left')
-_ = plt.title(f"Shallow classification tree with max-depth of {max_depth}")
+_ = plt.title(
+    f"Shallow classification tree with max-depth of {max_depth}"
+    f"\n Accuracy of the fit: {accuracy:.2f}"
+)
```
ogrisel, line 124: I don't see the point of measuring the scores only the training set. Here we speak about hyper-parameter tuning so this would be confusing to only display the training score. I think this notebook needs to be reworked to do a train test split and the plots should display both training and test errors, or neither.

Maybe the plots should be duplicated to each do 2 subplots: one with the prediction function displayed on top of a scatter plot of the samples of the training set (with the training score in the title) and another with the same prediction function displayed on top of a scatter plot of the samples of the testing set (with the testing score in the title).

And then we should comment on those scores to summarize the impact of the hyper-parameters in terms of the overfitting / underfitting trade-off.

ogrisel (commented):
I think this PR should be split because `python_scripts/trees_hyperparameters.py` needs more work as explained below while the rest should be good to go with the following small suggestion.

python_scripts/ensemble_hyperparameters.py
ogrisel:
```suggestion
print(f"On average, our GBDT regressor makes an error of {error:.2f} k$")
```

python_scripts/trees_hyperparameters.py
glemaitre: I split the PR.

PR #466, opened 2021-10-01
Add explanation on positive Ridge alpha

ArturoAmorQ:
This PR aims to give a short explanation on why Ridge `alpha` must be a positive number.

I also moved the paragraph out of the "Scale your data!" section as it did not seem to fit that topic.


python_scripts/linear_models_regularization.py
```diff
@@ -159,17 +159,18 @@
 # previous plot, we see that the magnitude of the weights are shrunk towards
 # zero in comparison with the linear regression model.
 #
+# Regularization will add constraints on weights of the model. We saw in the
```
glemaitre, line 162: Instead of this paragraph here, I would explicitly reproduce the previous experiment and increase even more alpha and call the attention of the user to the change of the scale of the weights and conclude regarding the impact of `alpha`. Once done, we can then add that a negative alpha does not make sense then.

glemaitre:
```python
# %% [markdown]
#
# In the previous example, we fixed `alpha=100`. We will now check the impact of the value
# of `alpha` by increasing its value.

# %%
# Add some code here to reproduce with alpha=1e6

# %% [markdown]
#
# Looking specifically to weights values, we observe that increasing the value of `alpha` 
# will decrease the weight values. A negative value of `alpha` would actually enhance the
# large weights and promote overfitting. Thus, in practice, `alpha` is always expected to be
# a positive value.
```

glemaitre (commented):
Otherwise I agree with adding more explanation

python_scripts/linear_models_regularization.py
```diff
@@ -231,9 +232,33 @@
...
-# Compare to the previous plots, we see that now all weight manitudes are
+# Compare to the previous plots, we see that now all weight magnitudes are
 # closer and that all weights are more equally contributing.
 #
+# In the previous example, we fixed `alpha=0.5`. We will now check the impact
+# of the value of `alpha` by increasing its value.
+
+# %%
+ridge = make_pipeline(PolynomialFeatures(degree=2), StandardScaler(),
+                      Ridge(alpha=500))
```
glemaitre, line 243: I would increase even more the regularization to bring the coefficients under 1

glemaitre (approved):
Otherwise LGTM

python_scripts/linear_models_regularization.py
```diff
@@ -163,24 +163,24 @@
...
 # parameter.
 #
-# ## Scale your data!
+# ## Feature scaling and regularization
 #
-# Regularization will add constraints on weights of the model. We saw in the
-# previous example that a ridge model will enforce that all weights have a
-# similar magnitude. Indeed, the larger alpha is, the larger this enforcement
-# will be.
+# On the one hand, weights provide a link between features and the target.
```
ogrisel, line 176:
```suggestion
# On the one hand, weights define the link between feature values and the
# predicted target.
```

python_scripts/linear_models_regularization.py
```diff
@@ -163,24 +163,24 @@
...
+# On the other hand, regularization adds constraints on the weights of the
+# model through the `alpha` parameter. Therefore, the effect that feature
+# rescaling has in all the weights also impacts the regularization.
```
ogrisel, line 179:
```suggestion
# rescaling has on the final weights also interacts with regularization.
```

python_scripts/linear_models_regularization.py
```diff
@@ -238,9 +238,33 @@
 _ = plt.title("Ridge weights with data scaling")
 
 # %% [markdown]
-# Compare to the previous plots, we see that now all weight manitudes are
+# Compare to the previous plots, we see that now all weight magnitudes are
 # closer and that all weights are more equally contributing.
```
ogrisel, line 242:
```suggestion
# closer and that all features are more equally contributing.
```

python_scripts/linear_models_regularization.py
```diff
@@ -333,6 +357,14 @@
 best_alphas = [est[-1].alpha_ for est in cv_results["estimator"]]
 best_alphas
 
+# %% [markdown]
+# The optimal regularization strength is not unique. In fact, there is a set of
+# models leading to equivalent generalization performances.
```
ogrisel, line 362:
```suggestion
# The optimal regularization strength is not necessarily the same on all
# cross-validation iterations. But since we expect each cross-validation
# resampling to stem from the same data distribution, it is common practice
# to use the average value of the best `alpha` found on different
# cross-validation folds as our final estimate for the tuned `alpha`.
```

ogrisel (approved):
Final pass of suggestions, but otherwise, LGTM. Will merge if CI is green.

PR #467, opened 2021-10-04
Add script to generate full index

### LINKED ISSUE #111: Create index.md to navigate notebooks while giving the course
We have more and more notebooks which makes it harder and harder to tell people open this notebook but not this one whose filename is almost the same. Also with long filenames in visioconference settings people don't see the full filename (in JupyterLab at least) and the filenames all start the same.

We should have a index.md file that allows to navigate between notebooks more easily. We would always come back to this index.md file at the end of each notebooks and say "OK we did this notebook and now we are going to do this one" click and off we go.

The binder link in the README could go to this index.

In an ideal world the index.md could be generated from the `_toc.yml` file but I think it would be OK to have it manually generated at first to test the idea.

lesteve:
Fixes #111.

This is useful for people that want to run the notebooks locally and also when giving courses that reuse the MOOC material so that people can more easily follow along.

I think manually creating indexes from the full one is the way to go (e.g. Inria Academy context).

### DIFF
--- a/build_tools/generate-md-index.py
+++ b/build_tools/generate-md-index.py
@@ -1,109 +1,18 @@
 # %%
 from pathlib import Path
-
-root_dir = Path(__file__).parents[1]
-# %%
-
 import yaml
-from pathlib import Path
 
-from sphinx_external_toc.parsing import parse_toc_yaml
-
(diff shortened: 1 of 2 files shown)


ogrisel:
LGTM, indeed this will be very useful for local teaching.

PR #473, opened 2021-10-27
MNT Erase unused line of code

ArturoAmorQ:
Tweak to remove line of code importing seaborn, as it is no longer necessary since  #250.


lesteve:
Thanks!

Going a bit more meta, it may be worth running `flake8` on all the `.py` files, ignoring some warnings (e.g. imports not at the top) and fixing them.

PR #477, opened 2021-11-03
MNT add justification for setting cv=5

ArturoAmorQ:
This PR adds a small paragraph to justify the use of either 5 or 10 folds in a cross-validation split


python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -390,9 +390,13 @@
 # trained to the full training set, along with an estimate of the variability
 # (uncertainty on the generalization accuracy).
 #
-# You can pass additional parameters to `cross_validate` to get more
-# information, for instance training scores. These features will be covered in
-# a future notebook.
+# Setting `cv=5` or `cv=10` is a common practice, as it is a good trade-off between
+# computation time and stability of the estimated variability.
```
glemaitre, line 394: It might be good to mention here that in the case we would like to obtain a distribution estimation of the scores then we usually would repeat this 5- or 10-fold cross-validation. This cross-validation is strategy is known as `RepeatedKFold` in scikit-learn and could be provided to the `cv` parameter.

glemaitre (commented):
I am not totally sure if this is the right moment to add a bit more information regarding the repeated k-fold but I think that it is nice to point out that `cv=int` is the easiest option but we can provide other scikit-learn cross-validation strategies that we will present later on in the MOOC.

python_scripts/02_numerical_pipeline_scaling.py
ArturoAmorQ: I guess it would be more appropriate to mention `RepeatedKFold` in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_train_test.html), as we introduce here the concept of score distributions. This could be a good occasion to extend the discussion on such a topic, what do you think?

ArturoAmorQ:
There was a paragraph with a bit redundant information that I rephrased to clarify that setting `cv=int` is equivalent to using the `KFold` strategy with `K` being said integer. Hope it is clearer now!

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -348,14 +348,14 @@
 # ![Cross-validation diagram](../figures/cross_validation_diagram.png)
```
glemaitre, line 348: We could as well have a colour legend if possible. This would be nice. It could be done in another PR.

glemaitre (approved):
LGTM as-is.

ogrisel (approved):
Some more comments, but otherwise, LGTM!

python_scripts/02_numerical_pipeline_scaling.py
```diff
@@ -348,14 +348,14 @@
...
 # ```{note}
-# This figure shows the particular case of K-fold cross-validation strategy.
-# As mentioned earlier, there are a variety of different cross-validation
-# strategies. Some of these aspects will be covered in more details in future
+# This figure shows the particular case of **K-fold** cross-validation strategy.
+# For each cross-validation split, the procedure trains a model on all the red
+# samples and evaluate the score of the model on the blue samples.
+# As mentioned earlier, there is a variety of different cross-validation
+# strategies. Some of these aspects will be covered in more detail in future
 # notebooks.
```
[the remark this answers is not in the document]
ogrisel: We are after the video if we follow: https://github.com/INRIA/scikit-learn-mooc/pull/492#issuecomment-988104295, so indeed, ignore this comment.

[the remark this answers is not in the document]
ArturoAmorQ: Ups! I commited too soon. I am reverting it

ogrisel (approved):

ogrisel:
Merged!

PR #480, opened 2021-11-05
FIX do not use `-` in pipeline step names 

### LINKED ISSUE #368: Avoid using dashes in pipeline step names
This avoids caveats when using set_params.

https://mooc-forums.inria.fr/moocsl/t/how-to-set-up-params-for-a-model/3513/2

Related to #364 that was werged and then reverted in #365 that was reverted because I was not sure whether there was an easy way to reset a CSV file to original in the JupyterHub user space. This is probably easier to do it for v2.

This should work to create a branch with the changes we want (reverting the commit that reverted the change is the same as actually doing the changes 😉)
```
git checkout -b no-dashes-in-pipeline-step-names
git revert fe03783 --no-commit
git commit -m 'Avoid using dashes in pipeline step names'
```

ArturoAmorQ:
Fixes #368. Now that the session of the MOOC has been closed for a while, it should be ok to edit the randomized_search_results_logistic_regression.csv file and re-revert the changes in pipelines' names.


glemaitre:
Thanks

PR #483, opened 2021-11-08
Replace plot_* function by Display in metrics_classification.py

### LINKED ISSUE #481: Replace plot_* function by Display
To avoid the deprecation warning, we should use the display instead of the `plot_*` functions

ArturoAmorQ:
Fixes #481. We were getting some deprecation warnings with scikit-learn >= 1 when using `plot_confusion_matrix`, `plot_precision_recall_curve` and `plot_roc_curve`.


glemaitre:
Thanks

PR #489, opened 2021-11-19
MNT update plot_decision_function by DecisionBoundaryDisplay

glemaitre:
Replace the `plot_decision_function` by a new display `DecisionBoundaryDisplay`.
We intend to add this display in `scikit-learn`. Thus, we place the import in a helper plotting module. In the future, only the import will be modified (together with a brief explanation regarding this class).

### DIFF
--- a/notebooks/helpers/plotting.py
+++ b/notebooks/helpers/plotting.py
@@ -1,0 +1,324 @@
+import numpy as np
+
+from sklearn.utils import check_matplotlib_support
+from sklearn.utils import _safe_indexing
+
+
+def _check_boundary_response_method(estimator, response_method):
+    """Return prediction method from the `response_method` for decision boundary.
+
+    Parameters
+    ----------
+    estimator : object
+        Estimator to check.
+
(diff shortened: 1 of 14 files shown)


python_scripts/helpers/plotting.py
```diff
@@ -0,0 +1,334 @@
+import numpy as np
```
ogrisel, line 1:
```suggestion
# This helper code is adapted from:
# https://github.com/scikit-learn/scikit-learn/pull/16061
# TODO: remove this code from the MOOC when the feature is
# made available in a stable version of scikit-learn itself.
import numpy as np
```

notebooks/helpers/plotting.py
```diff
@@ -0,0 +1,326 @@
+import numpy as np
```
ogrisel, line 1:
```suggestion
# This helper code is adapted from:
# https://github.com/scikit-learn/scikit-learn/pull/16061
# TODO: remove this code from the MOOC when the feature is
# made available in a stable version of scikit-learn itself.
import numpy as np
```

ogrisel:
Merged. thanks @glemaitre !

PR #496, opened 2021-12-07
MNT Improve consistency of definitions

ArturoAmorQ:
The [first exercise on linear models](linear_models_sol_01.py) uses the notation in singular `feature_name = "Flipper Length (mm)"` but some other notebooks define the variable in plural (i.e. `feature_names = "Flipper Length (mm)"`) even if there is one feature only. This PR solves this small detail and looks for consistency in notation across these notebooks.


ogrisel (approved):
LGTM!

Issue #75, opened 2020-10-26
Toggle navigation not working on pages with slide


lesteve:
https://inria.github.io/scikit-learn-mooc/linear_models/slides.html

Clicking on "Toggle navigation" icon (left arrow on the top left):
![image](https://user-images.githubusercontent.com/1680079/97154381-1754a300-1774-11eb-806e-33f65a0bc793.png)

does not do anything. It does work on any other page I tried e.g.https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models.html.

On the console:
```
Uncaught TypeError: $(...).tooltip is not a function
    initTooltips https://inria.github.io/scikit-learn-mooc/_static/sphinx-book-theme.be0a4a0c39cd630af62a2fcf693f3f06.js:8
    jQuery 9
    initTooltips https://inria.github.io/scikit-learn-mooc/_static/sphinx-book-theme.be0a4a0c39cd630af62a2fcf693f3f06.js:8
    sbRunWhenDOMLoaded https://inria.github.io/scikit-learn-mooc/_static/sphinx-book-theme.be0a4a0c39cd630af62a2fcf693f3f06.js:6
    <anonymous> https://inria.github.io/scikit-learn-mooc/_static/sphinx-book-theme.be0a4a0c39cd630af62a2fcf693f3f06.js:14
sphinx-book-theme.be0a4a0c39cd630af62a2fcf693f3f06.js:8:81
```

Maybe some conflict between the remark.js and the sphinx-book js?

Not so important, but in case someone has some suggestions.

GaelVaroquaux:
> Maybe some conflict between the remark.js and the sphinx-book js?

Unlikely: remark is in an iframe. It should be isolated

PR #497, opened 2021-12-10
Add YouTube hyperlinks to existing slides

ArturoAmorQ:
This PR aims to linking the different Youtube videos to their respective slides.

Please notice that the  iframe css in `sklearn_mooc.css` still has to be fixed in order to show the video and the slides at the same time. Do not merge this PR until that is solved.

Side-effect: fixes https://github.com/INRIA/scikit-learn-mooc/issues/75


ogrisel:
Hum, I think we need to fix the CSS before merging this because now the slides and video iframes cover each other...

ArturoAmorQ:
> Hum, I think we need to fix the CSS before merging this because now the slides and video iframes cover each other...

I agree. I will change to WIP.

lesteve:
(Side-comment since I was wondering about it, this does not affect FUN since they get the slides from `https://inria.github.io/scikit-learn-mooc/slides/?file=<file.md>` and not our JupyterBook .md file.

For example they use https://inria.github.io/scikit-learn-mooc/slides/?file=ml_concepts.md rather than https://inria.github.io/scikit-learn-mooc/ml_concepts/slides.html)

ogrisel:
This should be ready for review.

- I simplified the CSS to make the iframe use 100% of the width while keeping the 4/3 aspect ratio;
- Note that for some reasons, the `width="640px" height="480px"` attributes for the YouTube iframes are needed to get the video preview to render correctly, despite the actual width and height beeing overridden in the CSS;
- This also fixes the button to hide the left-side navigation panel which was annoying @lesteve when teaching live from the jupyterbook.

glemaitre (approved):
LGTM

Issue #449, opened 2021-09-01
Avoid using `permutation_test_score` to define chance level


ArturoAmorQ:
In [this lecture notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_baseline.html) and [this solution notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_sol_02) we define a notion of _chance level performance_ as a synonym of the `sklearn.model_selection.permutation_test_score`. This has two undesired consequences: 

- Implicitly introduces the concept of `pvalue`. Instead, we can do the shuffling by hand using `sklearn.utils.shuffle` to keep things conceptually simple.
- The concept of _chance level_ depends a lot of context. For the sake of clarity we can refer to this baseline as _permuted score/error_.

glemaitre:
> Implicitly introduces the concept of pvalue. Instead, we can do the shuffling by hand using sklearn.utils.shuffle to keep things conceptually simple.

I don't really see the problem here because we don't use or wither present the p-values. I would really prefer to use the functionality instead of doing the `for` loop and using the `sklearn.utils.shuffle`. I am not convinced that introducing a tool from `sklearn.utils` is a good idea.

> The concept of chance level depends a lot of context. For the sake of clarity we can refer to this baseline as permuted score/error.

I am fine changing the terminology to avoid any misunderstanding of concepts.

ogrisel:
FYI I agree with @ArturoAmorQ and I am working on a PR.

PR #499, opened 2021-12-13
Remove permutation test score

ogrisel:
Address #449 to avoid using `permutation_test_score` to define chance level as it was found to be confusing and it's arguably just one way to define a chance level, not "the" way.

The quizz is being adapted accordingly in the private repo:

https://gitlab.inria.fr/learninglab/mooc-scikit-learn/mooc-scikit-learn-coordination/-/merge_requests/32

The exercise instructions and the quizz instructions will be synced in a follow-up commit once those 2 PRs are reviewed and merged.

### DIFF
--- a/python_scripts/cross_validation_baseline.py
+++ b/python_scripts/cross_validation_baseline.py
@@ -1,23 +1,22 @@
 # %% [markdown]
 # # Comparing results with baseline and chance level
 #
-# In this notebook, we present how to compare the generalization performance of a
-# model to a minimal baseline.
+# In this notebook, we present how to compare the generalization performance of
+# a model to a minimal baseline.
 #
-# Indeed, in the previous notebook, we compared the testing error by
-# taking into account the target distribution. A good practice is to compare
-# the testing error with a dummy baseline and the chance level. In
-# regression, we could use the `DummyRegressor` and predict the mean target
-# without using the data. The chance level can be determined by permuting the
-# labels and check the difference of result.
+# Indeed, in the previous notebook, we compared the testing error by taking
+# into account the target distribution. A good practice is to compare the
+# testing error with a dummy baseline to define a chance level. In regression,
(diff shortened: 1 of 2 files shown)


python_scripts/cross_validation_sol_02.py
```diff
@@ -36,97 +37,127 @@
...
+test_score_logistic_regression
 
 # %% [markdown]
-# Finally, compute the test score of a dummy classifier which would predict
-# the most frequent class from the training set. You can look at the
-# `sklearn.dummy.DummyClassifier` class.
+# Now, compute the cross-validation scores of a dummy classifier that
+# constantly predicts the most frequent class observed the training set. Please
+# refer to the online documentation for the [sklearn.dummy.DummyClassifier
+# ](https://scikit-learn.org/stable/modules/generated/sklearn.dummy.DummyClassifier.html)
```
lesteve, line 60: Weird rendering here:
image

python_scripts/cross_validation_baseline.py
```diff
@@ -1,23 +1,18 @@
 # %% [markdown]
 # # Comparing results with baseline and chance level
```
lesteve, line 2: The title needs to be changed "Comparing results with baseline" or something similar since we don't mention chance level any more.

lesteve (commented):
Overall looks good to me, two minor comments.

python_scripts/cross_validation_baseline.py
ogrisel: Good points. Thanks for the review, I will push some fixes.

ogrisel:
Thanks for the review. The rendering looks good now. Once merged we will need to push another commit to update the exercise. Same for the private quiz solutions.

lesteve:
Merging, thanks!

I added a "FUN: requires changes" label since this changes a notebook title

PR #503, opened 2022-01-06
Use score in tree hyperpatameter notebook

glemaitre:
This PR isolate the call to score in the hyperparameter notebook.
It is linked to the comment: https://github.com/INRIA/scikit-learn-mooc/pull/464#discussion_r768893865

In this PR, we should therefore address the concern of @ogrisel:

I don't see the point of measuring the scores only the training set. Here we speak about hyper-parameter tuning so this would be confusing to only display the training score. I think this notebook needs to be reworked to do a train test split and the plots should display both training and test errors, or neither.

Maybe the plots should be duplicated to each do 2 subplots: one with the prediction function displayed on top of a scatter plot of the samples of the training set (with the training score in the title) and another with the same prediction function displayed on top of a scatter plot of the samples of the testing set (with the testing score in the title).

And then we should comment on those scores to summarize the impact of the hyper-parameters in terms of the overfitting / underfitting trade-off.

### DIFF
--- a/python_scripts/trees_hyperparameters.py
+++ b/python_scripts/trees_hyperparameters.py
@@ -114,13 +114,23 @@ tree_reg = DecisionTreeRegressor(max_depth=max_depth)
 # %%
 plot_classification(tree_clf, data_clf[data_clf_columns],
                     data_clf[target_clf_column])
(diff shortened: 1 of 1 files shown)


ogrisel (commented):
I started to review this PR but I noticed that it requires blackification first.

PR #506, opened 2022-01-06
Improve group identification in digits example

glemaitre:
We have a small discrepancy between the number of writers identified in the target and the one provided by the documentation. However, we are able to show the effect of the groups. So we provide a small discussion just to avoid surprises.

### DIFF
--- a/python_scripts/cross_validation_grouping.py
+++ b/python_scripts/cross_validation_grouping.py
@@ -15,11 +15,11 @@ data, target = digits.data, digits.target
 # a logistic regression classifier with preprocessor to scale the data.
 
 # %%
-from sklearn.preprocessing import StandardScaler
+from sklearn.preprocessing import MinMaxScaler
 from sklearn.linear_model import LogisticRegression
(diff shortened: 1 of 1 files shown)


ogrisel (approved):

lesteve:
Thanks merging!

python_scripts/cross_validation_grouping.py
```diff
@@ -105,7 +105,23 @@
...
+#
+# It might not be obvious at first, but there is a structure in the target:
+# there is a repetitive pattern that always starts by some series of ordered
+# digits from 0 to 9 followed by random digits at a certain point. If we look
+# in details, we see that there is 14 such patterns, always with around 130
+# samples each.
+#
+# Even if it is not exactly corresponding to the 13 writers in the
+# documentation (maybe one writer wrote two series of digits), we can
+# make the hypothesis that each of these structure corresponds to a different
```
ogrisel, line 123:
```suggestion
# make the hypothesis that each of these structures corresponds to a different
```

PR #508, opened 2022-01-06
Fix wording in HGBDT notebook

### LINKED ISSUE #443: KBinsDiscretizer drastically reduced fit time
This [post in the forum](https://mooc-forums.inria.fr/moocsl/t/discretizer-drastically-reduced-fit-time/4237) complains about the use of the word _drastically_ in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_hist_gradient_boosting.html) for describing the improvement of using `KBinsDiscretizer` along with `GradientBoostingRegressor`. 

That paragraph could be rephrased not only to suppress the word _drastically_ but also to include [Olivier's detailed answer](https://mooc-forums.inria.fr/moocsl/t/discretizer-drastically-reduced-fit-time/4237/5) on the `nlog(n)` vs. linear complexity when introducing `HistGradientBoostingRegressor`.

What do you think?

glemaitre:
closes #443 

Remove "drastically" and rephrase the part about integer values.

### DIFF
--- a/python_scripts/ensemble_hist_gradient_boosting.py
+++ b/python_scripts/ensemble_hist_gradient_boosting.py
@@ -85,7 +85,7 @@ data_trans
 # for those features. The smallest bins will be removed.
 # ```
 # We see that the discretizer transforms the original data into integral
-# values (eventhough they are encoded using a floating-point precision).
+# values (eventhough they are encoded using a floating-point representation).
 # Each value represents the bin index when the distribution by quantile is
 # performed. We can check the number of bins per feature.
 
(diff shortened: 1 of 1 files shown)


python_scripts/ensemble_hist_gradient_boosting.py
```diff
@@ -84,8 +84,9 @@
 # the features, we requested too much bins in regard of the data dispersion
 # for those features. The smallest bins will be removed.
 # ```
-# We see that the discretizer transforms the original data into an integer.
-# This integer represents the bin index when the distribution by quantile is
+# We see that the discretizer transforms the original data into integral
+# values (eventhough they are encoded using a floating-point precision).
```
ogrisel, line 88:
```suggestion
# values (eventhough they are encoded using a floating-point representation).
```

python_scripts/ensemble_hist_gradient_boosting.py
```diff
@@ -117,7 +118,7 @@
       f"{cv_results_gbdt['score_time'].mean():.3f} seconds")
 
 # %% [markdown]
-# Here, we see that the fit time has been drastically reduced but that the
+# Here, we see that the fit time has been reduced but that the
 # generalization performance of the model is identical. Scikit-learn provides a
 # specific classes which are even more optimized for large dataset, called
```
ogrisel, line 123:
```suggestion
# generalization performance of the model is identical. Scikit-learn provides
# specific classes which are even more optimized for large dataset, called
```

ogrisel (approved):
LGTM once the following is dealt with:

PR #509, opened 2022-01-06
Add note advising for error_score='raise' to debug

### LINKED ISSUE #377: Add a technical note regarding the parameter `error_score` in cross-validation tools
Usually, users are not digging enough to get why they obtain `NaN` with cross-validation tools (e.g. `cross_validate`). Indeed, we raise a warning but most probably, they just ignore it since this is not an error and just focus on the `NaN`.

I am usually posting that they can debug the code with `error_score="raise"` and I think that we should add this note somewhere with a concrete example.

glemaitre:
closes #377

### DIFF
--- a/python_scripts/03_categorical_pipeline_ex_01.py
+++ b/python_scripts/03_categorical_pipeline_ex_01.py
@@ -75,10 +75,13 @@ from sklearn.linear_model import LogisticRegression
 #
 # ```{note}
 # Be aware that if an error happened during the cross-validation,
-# `cross_validate` will raise a warning and will return NaN as scores. You can
-#  get instead a standard Python traceback, you can set the parameter
-# `error_score="raise"` in `cross_validate` that will raise the first exception
-# encountered. It is particularly handy when debugging your complex pipelines.
+# `cross_validate` will raise a warning and return NaN (Not a Number)
+# as scores.  To make it raise a standard Python exception with a traceback,
+# you can pass the `error_score="raise"` argument in the call to
+# `cross_validate`. An exception will be raised instead of a warning at the first
+# encountered problem  and `cross_validate` will stop right away instead of
+# returning NaN values. This is particularly handy when developing
+# complex machine learning pipelines.
 # ```
 
 # %%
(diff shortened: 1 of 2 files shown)


python_scripts/03_categorical_pipeline_ex_01.py
```diff
@@ -72,6 +72,14 @@
...
 # Your model is now defined. Evaluate it using a cross-validation using
 # `sklearn.model_selection.cross_validate`.
+#
+# ```{note}
+# Be aware that if an error happened during the cross-validation,
+# `cross_validate` will raise a warning and will return NaN as scores. You can
+#  get instead a standard Python traceback, you can set the parameter
+# `error_score="raise"` in `cross_validate` that will raise the first exception
+# encountered. It is particularly handy when debugging your complex pipelines.
+# ```
```
ogrisel, line 82:
```suggestion
# ```{note}
# Be aware that if an error happened during the cross-validation,
# `cross_validate` will raise a warning and return NaN (Not a Number)
# as scores.  To make it raise a standard Python exception with a traceback,
# you can pass the `error_score="raise"` argument in the call to
# `cross_validate`. An exception will be raised instead of a warning at the first
# encountered problem  and `cross_validate` will stop right away instead of
# returning NaN values. This is particularly handy when developing
# complex machine learning pipelines.
# ```
```

python_scripts/03_categorical_pipeline_sol_01.py
```diff
@@ -74,6 +74,14 @@
...
 # Your model is now defined. Evaluate it using a cross-validation using
 # `sklearn.model_selection.cross_validate`.
+#
+# ```{note}
+# Be aware that if an error happened during the cross-validation,
+# `cross_validate` will raise a warning and will return NaN as scores. You can
+#  get instead a standard Python traceback, you can set the parameter
+# `error_score="raise"` in `cross_validate` that will raise the first exception
+# encountered. It is particularly handy when debugging your complex pipelines.
+# ```
```
ogrisel, line 84:
```suggestion
# ```{note}
# Be aware that if an error happened during the cross-validation,
# `cross_validate` will raise a warning and return NaN (Not a Number)
# as scores.  To make it raise a standard Python exception with a traceback,
# you can pass the `error_score="raise"` argument in the call to
# `cross_validate`. An exception will be raised instead of a warning at the first
# encountered problem  and `cross_validate` will stop right away instead of
# returning NaN values. This is particularly handy when developing
# complex machine learning pipelines.
# ```
```
```

ogrisel (commented):
Suggested re-phrasing:

PR #512, opened 2022-01-06
Remove AdaBoost from exercises in boosting

### LINKED ISSUE #456: Edit Exercise M6.03 on Adaboost overfitting
As mentioned in [Olivier's comment](https://mooc-forums.inria.fr/moocsl/t/quiz-m6-03-question-3/3128/15):

> when increasing the number of trees in AdaBoost on this dataset (California housing), both the training and validation errors increase and the average validation error is close to the average training error, meaning that overfitting is not the main effect but underfitting is counter-intuitively increasing with the number of estimators.

We decided to keep the [Adaptive Boosting (AdaBoost) notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_adaboost.html) as it is, but we agreed that the [Exercise M6.03 notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_ex_03.html) and [its solution](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_03.html) should be edited somehow.

Removing the AdaBoost part of the exercise and just keeping the validation curve for the RandomForest would make this notebook more suitable to the section "Ensemble method using bootstrapping". 

Whatever we decide to do, we have to edit Quiz M6.03 Q3 accordingly. The subject is still open to discussion.

glemaitre:
closes #456 

Remove exercise that play with AdaBoost and move learning curve into a more general boosting exercise.


lesteve:
First thing: I think the impact on FUN is reduced since we are removing Exercise M6.04 which is the last one of the module, there is no need to rename plenty of exercises. For FUN, this is "only" a reordering: the "Gradient-boosting decision tree (GBDT)" notebook needs to be moved before "Exercise M6.03".

lesteve:
For some reason there is no rendering of the outputs in the artifact HTML: https://2577-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/ensemble_sol_03.html, not sure why ...

ogrisel (approved):
LGTM!

ogrisel:
> For some reason there is no rendering of the outputs in the artifact HTML: https://2577-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/ensemble_sol_03.html, not sure why ...

This is annoying...

Issue #276, opened 2021-04-20
Feedback on Linear Model Chapter


lorentzenchr:
This MOOC is really great. Below are a few suggestions after a first partial read.

Chapter [Linear regression using scikit-learn](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_regression_in_sklearn.html):

- Mention that the strategy for finding the best parameters is called (ordinary) **least squares**.

[Linear regression with non-linear link between data and target](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_regression_non_linear_link.html)

- The word *link* is unfortunate as it might get confused with the *link* function of GLMs.
- As soon as v1.0 of scikit-learn arrives, mention splines with `SplineTransformer`.
- Speeking about *interactions* is a bit misleading in a setting with only one single feature. But I would really love to see an example that specifies 2 interaction terms: Given x1 to x3, create x1, x2, x3, x1*x2, x1*x3 (not x2*x3 and not x**2 terms either).

[Regularization of linear regression model](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_regularization.html)

- Remove convergence warnings of Ridge (other solver?)
- Maybe a word about categorical variables and the `drop` parameter in OHE?
- "Let’s consider the case where features have an identical data dispersion". The term "dispersion" seems unusual to me. How about "features live on the same scale/dimension like both are times measured in microseconds."
   Later on: "our model will boost the weights of features with small dispersion" => It is about larger/smaller (floating point/machine) numbers, not about dispersion.
- What about scaling of OHE categorical features?

[Main take-away](https://inria.github.io/scikit-learn-mooc/linear_models/linear_models_module_take_away.html)

- Again the word "link".
- Mention that (manual) feature engineering and feature selection is (by far) the greatest impact/tuning factor for linear models (and therefore against over- and underfitting).

lesteve:
> This MOOC is really great. Below are a few suggestions after a first partial read.

Thanks for the kind words! It's great to hear, especially since we have a beta-test phase going right now and we are hearing about many many small things that could be better.

Thanks for taking the time to put together some feed-back as well, this is really valuable! Working on the MOOC I realized that reviewing pedagogical content is even harder than reviewing code and reviewing code is already quite hard.

Looking at the points you raised, I tried to categorise into different buckets. Let me know if you have any comments on this rough categorisation!

### Things I think we should aim to fix reasonably quickly (end-April?)
[...]

PR #514, opened 2022-01-06
Use scale instead of dispersion

glemaitre:
Partially address #276 by removing "dispersion" and using "scale" for this specific case.


ogrisel (approved):
LGTM.

PR #515, opened 2022-01-06
Merge M4.04a and M4.04b quizzes

### LINKED ISSUE #469: Merge M4.04a and M4.04b quizzes
This happened when moving parts around after splitting videos. There is a single question in the M4.04a quiz and is the only part in the MOOC where we have two quizzes in the same lesson (everywhere else it is one quizz per lesson at the end of the lesson).

Alternatively we could add more questions about the regularisation video (this is more work).

lesteve:
Fixes #469 

Follow-up of gitlab MR: https://gitlab.inria.fr/learninglab/mooc-scikit-learn/mooc-scikit-learn-coordination/-/merge_requests/35


ogrisel (approved):
LGTM!

PR #516, opened 2022-01-06
Add a note for scaling categorical features with linear models

glemaitre:
partially address #276 

Discuss that scaling is usually omitted with `OneHotEncoder`


python_scripts/linear_models_regularization.py
```diff
@@ -270,6 +272,55 @@
...
+categories_encoded = pd.DataFrame(
+    categories_encoded,
+    columns=encoder.get_feature_names_out(input_features=["marital_status"]),
+)
+categories_encoded
+
+# %% [markdown]
+#
 # In the next section, we will check the impact of this hyperparameter and how
 # it should be tuned.
```
ogrisel, line 325:
```suggestion
# In the next section, we will check the impact of the regularization
# parameter `alpha `and how it should be tuned.
```

python_scripts/linear_models_regularization.py
```diff
@@ -270,6 +272,55 @@
...
+
+# %%
+import numpy as np
+from sklearn.preprocessing import OneHotEncoder
+
+categories = np.array(
+    ["married"] * 5 + ["not_married"] * 5, dtype=object
+)[:, np.newaxis]
+
+encoder = OneHotEncoder(sparse=False, dtype=int)
```
ogrisel, line 297: It's a bad practice to pass platform specific dtypes to numpy.

```suggestion
encoder = OneHotEncoder(sparse=False, dtype=np.int32)
```

python_scripts/linear_models_regularization.py
```diff
@@ -270,6 +272,55 @@
...
+# %% [markdown]
+#
+# At the beginning of this notebook, we saw that one should avoid such
+# collinearity when fitting a linear model.
+#
+# The `OneHotEncoder` provides a parameter `drop` that allows to drop a
+# the first feature (always or only in the binary case).
+
+# %%
+encoder = OneHotEncoder(drop="first", sparse=False, dtype=int)
```
ogrisel, line 314: same comment here.

ogrisel (commented):
I find it a bit weird to insert this comment in the middle of this notebook. I think it breaks the pedagogical progression. Furthermore we do not actually fit a model on the data pre-processed this way in the rest of the notebook, so it's a bit awkward to recommend something and then not do it afterwards.

Maybe moving this note to the end of the notebook, would be better?

python_scripts/linear_models_regularization.py
```diff
@@ -270,6 +272,55 @@
...
+    columns=encoder.get_feature_names_out(input_features=["marital_status"]),
+)
+categories_encoded
+
+# %% [markdown]
+#
+# At the beginning of this notebook, we saw that one should avoid such
+# collinearity when fitting a linear model.
+#
+# The `OneHotEncoder` provides a parameter `drop` that allows to drop a
```
ArturoAmorQ, line 310:
```suggestion
# The `OneHotEncoder` provides a parameter `drop` that allows to drop
```

glemaitre:
> Maybe moving this note to the end of the notebook, would be better?

Yep let me check how to improve that. I was thinking that being part of scaling was fine but actually with `drop` it is not only about scaling anymore.

python_scripts/linear_models_regularization.py
```diff
@@ -266,12 +266,25 @@
...
+# ```
+#
 # In the previous analysis, we did not study if the parameter `alpha` will have
 # an effect on the performance. We chose the parameter beforehand and fix it
 # for the analysis.
 #
-# In the next section, we will check the impact of this hyperparameter and how
-# it should be tuned.
+# In the next section, we will check the impact of the regularization
+# parameter `alpha `and how it should be tuned.
```
ArturoAmorQ, line 287:
```suggestion
# parameter `alpha` and how it should be tuned.
```

python_scripts/linear_models_regularization.py
```diff
@@ -266,12 +266,25 @@
 # of `alpha` will decrease the weight values. A negative value of `alpha` would
 # actually enhance large weights and promote overfitting.
 #
+# ```{note}
+# Here, we only focus on numerical features. For categorical features, it is
+# generally common to omit scaling when features are encoded with a
+# `OneHotEncoder` since the feature values are already on a similar scale.
+#
+# However, there is no good practice since scaling has an interaction with the
```
ogrisel, line 274:
```suggestion
# However, this choice can be questioned since scaling interacts with
```

python_scripts/linear_models_regularization.py
```diff
@@ -266,12 +266,25 @@
...
+# regularization as well. For instance, scaling categorical features that are
+# imbalanced (e.g. more occurrences of a specific category) allows to apply a
+# similar penalty to each of the category. However, scaling such features in
+# presence of rare category could be problematic (i.e. division by a very
+# small std. dev.) and it will introduce some numerical issue.
```
ogrisel, line 279:
```suggestion
# imbalanced (e.g. more occurrences of a specific category) would even out
# the impact of regularization to each category. However, scaling such features
# in the presence of rare categories could be problematic (i.e. division by a very
# small standard deviation) and it can therefore introduce numerical issues.
```

ogrisel (approved):

PR #520, opened 2022-01-07
Renaming master to main

ogrisel:
I did a `git grep master` to find them all.

I also bump-up the cache key of github actions.


glemaitre (approved):
It looks good. Let see if the CI passes

glemaitre:
It seems that it works and that the site is available: https://2609-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/01_tabular_data_exploration.html

ogrisel:
Yes but the `jupyter-book/_changed.html` summary file is missing.

I will try to investigate.

PR #522, opened 2022-01-07
Add information regarding the drop option in the linear model exercise

### LINKED ISSUE #276: Feedback on Linear Model Chapter

glemaitre:
closes #276 

Introduce the parameter `drop` from the `OneHotEncoder` through the exercise presenting the problem of collinear features.

### DIFF
--- a/python_scripts/linear_models_sol_04.py
+++ b/python_scripts/linear_models_sol_04.py
@@ -98,10 +98,10 @@ feature_names = [
     "Noisy feature #0",
     "Noisy feature #1",
     "Noisy feature #2",
-    "Repeated feature #0",
-    "Repeated feature #1",
-    "Repeated feature #0",
-    "Repeated feature #1",
+    "First repetition of feature #0",
+    "First repetition of  feature #1",
+    "Second repetition of  feature #0",
+    "Second repetition of  feature #1",
 ]
 coef = pd.Series(linear_regression.coef_, index=feature_names)
 _ = coef.plot.barh()
@@ -218,10 +218,10 @@ X_trans
 # earlier in this exercise. Training a linear regression model on such a
 # of one-hot encoded binary feature can therefore lead to numerical
 # problems, especially without regularization. Furthermore, the two one-hot
-# feature are redundant as they encode exactly the same information in
+# features are redundant as they encode exactly the same information in
 # opposite ways.
 #
-# Using regularization will helps to overcome the numerical issues that we highlighted
+# Using regularization helps to overcome the numerical issues that we highlighted
 # earlier in this exercise.
 #
 # Another strategy is to arbitrarily drop one of the encoded categories.
--- a/python_scripts/linear_models_sol_04.py
+++ b/python_scripts/linear_models_sol_04.py
@@ -30,7 +30,13 @@ data, target, coef = make_regression(
 # %%
 import pandas as pd
 
-feature_names = [f"Features {i}" for i in range(data.shape[1])]
+feature_names = [
+    "Relevant feature #0",
+    "Relevant feature #1",
+    "Noisy feature #0",
+    "Noisy feature #1",
+    "Noisy feature #2",
+]
 coef = pd.Series(coef, index=feature_names)
 coef.plot.barh()
 coef
@@ -49,7 +55,13 @@ linear_regression.fit(data, target)
 linear_regression.coef_
 
 # %% tags=["solution"]
-feature_names = [f"Features {i}" for i in range(data.shape[1])]
+feature_names = [
+    "Relevant feature #0",
+    "Relevant feature #1",
+    "Noisy feature #0",
+    "Noisy feature #1",
+    "Noisy feature #2",
+]
 coef = pd.Series(linear_regression.coef_, index=feature_names)
 _ = coef.plot.barh()
 
@@ -80,7 +92,17 @@ linear_regression.fit(data, target)
 linear_regression.coef_
 
 # %% tags=["solution"]
-feature_names = [f"Features {i}" for i in range(data.shape[1])]
+feature_names = [
+    "Relevant feature #0",
+    "Relevant feature #1",
+    "Noisy feature #0",
+    "Noisy feature #1",
+    "Noisy feature #2",
+    "Repeated feature #0",
+    "Repeated feature #1",
+    "Repeated feature #0",
+    "Repeated feature #1",
+]
 coef = pd.Series(linear_regression.coef_, index=feature_names)
 _ = coef.plot.barh()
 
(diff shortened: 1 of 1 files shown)


python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +136,101 @@
...
+    X_trans,
+    columns=encoder.get_feature_names_out(input_features=single_feature),
+)
+X_trans
+
+# %% [markdown] tags=["solution"]
+#
+# Here, we see that the encoded category "CentralAir_N" is the opposite of the
+# encoded category "CentralAir_Y". Therefore, we observe that using a
+# `OneHotEncder` creates features having the problematic pattern observe
```
ogrisel, line 191:
```suggestion
# `OneHotEncoder` creates two features having the problematic pattern observed
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +136,101 @@
...
+# earlier in this exercise. Training a linear model, only one of the feature
+# is therefore informative because they carry the same information.
```
ogrisel, line 193:
```suggestion
# earlier in this exercise. Training a linear regression model on such a
# of one-hot encoded binary feature can therefore lead to numerical
# problems, especially without regularization. Furthermore, the two one-hot
# feature are redundant as they encode exactly the same information in
# opposite ways.
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +136,101 @@
...
+#
+# Using regularization will thus help to overcome the issues that we spotted
```
ogrisel, line 195:
```suggestion
# Using regularization will helps to overcome the numerical issues that we highlighted
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +136,101 @@
...
+# earlier in this exercise.
+#
+# Another strategy is to arbitrary drop one of the encoded categories.
```
ogrisel, line 198:
```suggestion
# Another strategy is to arbitrarily drop one of the encoded categories.
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +136,101 @@
...
+    columns=encoder.get_feature_names_out(input_features=single_feature),
+)
+X_trans
+
+# %% [markdown] tags=["solution"]
+#
+# We see that only the second column of the previous encoded data is kept.
+# This preprocessing is commonly used. However, you should notice that it
+# break feature symmetry and will impact the number of coefficients of the
+# model, their values, and thus their meaning.
```
ogrisel, line 219:
```suggestion
# Dropping one of the one-hot encoded column is a common practice,
# especially for binary categorical features. Note however that this breaks
# symmetry between categories and impacts the number of coefficients of the
# model, their values, and thus their meaning, especially when applying
# strong regularization.
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +136,101 @@
...
+#
+# Note that we can use this option is a machine-learning pipeline.
```
ogrisel, line 221:
```suggestion
# Let's finally illustrate how to use this option is a machine-learning pipeline:
```

glemaitre:
Let me also rename the feature names as we previously discussed during a call to make it easier to read the coefficients.

python_scripts/linear_models_sol_04.py
```diff
@@ -80,7 +92,17 @@
...
+feature_names = [
+    "Relevant feature #0",
+    "Relevant feature #1",
+    "Noisy feature #0",
+    "Noisy feature #1",
+    "Noisy feature #2",
+    "Repeated feature #0",
+    "Repeated feature #1",
+    "Repeated feature #0",
+    "Repeated feature #1",
```
ogrisel, line 104:
```suggestion
    "First repetition of feature #0",
    "First repetition of  feature #1",
    "Second repetition of  feature #0",
    "Second repetition of  feature #1",
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +158,110 @@
...
+
+# %% [markdown] tags=["solution"]
+#
+# Here, we see that the encoded category "CentralAir_N" is the opposite of the
+# encoded category "CentralAir_Y". Therefore, we observe that using a
+# `OneHotEncoder` creates two features having the problematic pattern observed
+# earlier in this exercise. Training a linear regression model on such a
+# of one-hot encoded binary feature can therefore lead to numerical
+# problems, especially without regularization. Furthermore, the two one-hot
+# feature are redundant as they encode exactly the same information in
```
ogrisel, line 221:
```suggestion
# features are redundant as they encode exactly the same information in
```

python_scripts/linear_models_sol_04.py
```diff
@@ -136,3 +158,110 @@
...
+# opposite ways.
+#
+# Using regularization will helps to overcome the numerical issues that we highlighted
```
ogrisel, line 224:
```suggestion
# Using regularization helps to overcome the numerical issues that we highlighted
```

ogrisel (approved):
LGTM with the following remaining details. I will commit an merge.

PR #524, opened 2022-01-07
add local datasets for financial data

### LINKED ISSUE #507: Check that code can run without internet access
In future editions we might want to configure the jupyterhub cluster to disable internet access for the notebook kernels.

It means in particular that we should not have notebooks that call `pd.read_csv` on `https://` URL to load some datasets for instance.

We have one such example in:

- https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_time.html

There might be others.

glemaitre:
closes #507 

Add a local version of the financial data for the cross-validation non-iid exercice.


ogrisel (approved):
Thanks!

PR #525, opened 2022-01-07
use DecisionBoundaryDisplay in tree hyperparameter

### LINKED ISSUE #510: Can we use plotting helpers in decision tree hyperparameter tuning
This uses complicated code to decision functions:
https://inria.github.io/scikit-learn-mooc/python_scripts/trees_hyperparameters.html#create-helper-functions

Can the plotting helpers be used instead? Plotting helpers are here:
https://github.com/INRIA/scikit-learn-mooc/blob/master/python_scripts/helpers/plotting.py

glemaitre:
closes #510 

Use the display to simplify the huge amount of matplotlib code in the notebook.

### DIFF
--- a/python_scripts/trees_hyperparameters.py
+++ b/python_scripts/trees_hyperparameters.py
@@ -26,10 +26,10 @@ data_reg = pd.read_csv("../datasets/penguins_regression.csv")
 # ```
 
 # %% [markdown]
-# ## Create helper function
+# ## Create helper functions
 #
-# We will create a small helper function to plot the results of the regression
-# trees.
+# We will create some helper functions to plot the data samples as well as the
+# decision boundary for classification and the regression line for regression.
 
 # %%
 import numpy as np
--- a/python_scripts/trees_hyperparameters.py
+++ b/python_scripts/trees_hyperparameters.py
@@ -38,7 +38,8 @@ import seaborn as sns
 from helpers.plotting import DecisionBoundaryDisplay
 
 
-def plot_classification(model, data, feature_names, target_names):
+def fit_and_plot_classification(model, data, feature_names, target_names):
+    model.fit(data[feature_names], data[target_names])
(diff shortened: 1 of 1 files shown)


python_scripts/trees_hyperparameters.py
```diff
@@ -112,12 +67,23 @@ def plot_regression(model, X, y, ax=None):
...
+tree_clf = DecisionTreeClassifier(max_depth=max_depth).fit(
+    data_clf[data_clf_columns], data_clf[target_clf_column]
+)
+tree_reg = DecisionTreeRegressor(max_depth=max_depth).fit(
+    data_reg[data_reg_columns], data_reg[target_reg_column]
+)
 
 # %%
-plot_classification(tree_clf, data_clf[data_clf_columns],
-                    data_clf[target_clf_column])
```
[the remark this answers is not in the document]
glemaitre: This is fine with me. Like this, `plot_regression` and `plot_classification` will do the same things.

python_scripts/trees_hyperparameters.py
```diff
@@ -28,73 +28,39 @@
...
-    Z = Z.reshape(xx.shape)
-
-    # make the plot of the boundary and the data samples
-    if ax is None:
-        _, ax = plt.subplots()
-    ax.contourf(xx, yy, Z, alpha=0.4, cmap="RdBu")
-    if y.nunique() == 3:
-        palette = ["tab:red", "tab:blue", "black"]
-    else:
+def plot_classification(model, data, feature_names, target_names):
```
ogrisel, line 41: Maybe you could rename the function to `fit_and_plot_classification` and actually do the call to fit in the function?

glemaitre: Since we always filtering the columns and target, I think this is better to include it in the function.

ogrisel (approved):

PR #528, opened 2022-01-10
Avoid reference to `l2` when talking about regularization

### LINKED ISSUE #451: Avoid use of l2 penalty/l2 regularization
Even if we have a section _Regularization in linear model_ where we present  `Ridge`'s alpha and `LogisticRegression`'s C to introduce the concept of regularizing parameters, we don't really go into the mathematical background of l1 and l2 regularizations. We may want to avoid introducing concepts to keep things as simple as possible.

Variants of the term l2 penalty/l2 regularization are used in the following notebooks:

- [Hyperparameter tuning by randomized-search](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_randomized_search.html) referring to the hyperparameter in `HistGradientBoostingClassifier`. This is fine as it is regarded as another hyperparameter and a simple explanation is provided. 
- [Quiz M3.02](https://inria.github.io/scikit-learn-mooc/tuning/parameter_tuning_automated_quiz_m3_02.html) also referring to the hyperparameter in `HistGradientBoostingClassifier`. This is also fine as it is the sequence of the randomized-search notebook.
- [Solution for Exercise M4.04](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_sol_04.html) mentioned inside a _Tip message_ referring to the general use of regularizing parameters such as `Ridge` alpha and `LogisticRegression` C. This can easily be rephrased.
- [Quiz M4.04b](https://inria.github.io/scikit-learn-mooc/linear_models/linear_models_quiz_m4_04b.html) referring to Ridge alpha. Notice that Q1 and Q5 are too similar. We could merge them and use simply "alpha" instead.
- [Exercise M4.05](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_ex_05.html) and [its solution](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_sol_05.html) set `penalty="l2"` in the argument of the `LogisticRegression`. This can be suppressed as it is already the default penalty for this classifier.

- [Quiz M4.05](https://inria.github.io/scikit-learn-mooc/linear_models/linear_models_quiz_m4_05.html) asks if the default setting of the `LogisticRegression` is a) no penalty;  b) a l2 penalty;  c) a l1 penalty. We can appeal to the C parameter only, has the l1 penalty is not mentioned elsewhere in the MOOC.

As elsewhere in the MOOC we either use alpha or C, accordingly, I suggest we change l2-regularization -> regularization and avoid appealing to "l2" before an hypothetical module on feature selection. What do you think?

**Edit:** Another option (as discussed with Guillaume) is to add a small description explaining that both alpha (for `RidgeCV`) and  C (for `LogisticRegression`) are examples of a broader category of hyperparameters.

ArturoAmorQ:
Closes #451.

The problem is about mentioning "l2 regularization" instead of just "regularization" since we do not explain what "l2" stands for and we don't want to do it before an hypothetical module on feature selection.

**Note to ourselves:** This PR has to be merged along with the respective gitlab MR.


jupyter-book/linear_models/linear_models_quiz_m4_05.md
```diff
@@ -4,8 +4,8 @@
 By default, a `LogisticRegression` in scikit-learn applies:
 
 - a) no penalty
-- b) a l2 penalty
-- c) a l1 penalty
+- b) a penalty whose strength can be tuned by the user
+- c) a cross-validation to select the penalty strength
```
glemaitre, line 8: I still would prefer to have a question linked to the penalty here. Indeed, the documentation in scikit-learn provide "none", "l1", "l2" penalty as parameter. So I would propose to:

- add a link to the documentation to make the quiz easier (basically a hyperlink from `LogisticRegression` to the doc)
- add the following solution that is more explicit and linked with the course:

```
- b) shrunk the magnitude of the weights toward zero (also called "l2 penalty")
- c) set some weights exactly to zero (also called "l1 penalty")
```

I think that we wanted people to discover (or make sure that they are aware) that `LogisticRegression` has a default penalty by default.

python_scripts/linear_models_sol_04.py
```diff
@@ -131,7 +131,7 @@
 
 # %% [markdown] tags=["solution"]
 # ```{tip}
-# We always advise to use l2-penalized model instead of non-penalized model
+# We advise to always use penalized models instead of non-penalized models
```
glemaitre, line 134: I don't think this is the right message now. We exactly want to advise using a l2 penalty. E.g. we don't want people to use a l1 penalty. So we can make it explicit (as in my previous quiz comment) what type of penalty (shrink weight magnitude towards zeros) and add it explicitly that this is called l2 penalty.

python_scripts/linear_models_sol_05.py
```diff
@@ -1,11 +1,12 @@
...
-# In the previous notebook, we presented a non-penalized logistic regression
-# classifier. This classifier accepts a parameter `penalty` to add a
-# regularization. The regularization strength is set using the parameter `C`.
-#
-# In this exercise, we ask you to train a l2-penalized logistic regression
-# classifier and to find by yourself the effect of the parameter `C`.
+# In the previous notebook we presented the `penalty` parameter of a logistic
+# regression classifier. It controls the **type** of regularization to use,
+# whereas the regularization **strength** is set using the parameter `C`. In
+# that case we set `penalty="none"`, which is equivalent to an infinitely large
```
glemaitre, line 6: In the example, you are not setting the penalty, so by default, you have a `"l2"` penalty which is the gotcha here. I would still force to use a l2 penalty. We can be explicit that we want the same penalty than in Ridge that shrinks weight magnitude towards zero and that is called l2.

ogrisel:
Hum I concurrently merged the MR in the private repo with the quiz solutions without taking this review into account. Next time let's not a open a PR on the quiz files in the public repo but only an MR on the private repo to review questions and answers together.

Once the MR is merged in the private repo, just run `make quizzes` on the public repo (with updated repo) and push to `main` directly without doing a PR.

python_scripts/linear_models_sol_05.py
ArturoAmorQ: Done, thanks!

ArturoAmorQ:
> Hum I concurrently merged the MR in the private repo with the quiz solutions without taking this review into account. Next time let's not a open a PR on the quiz files in the public repo but only an MR on the private repo to review questions and answers together.

Don't worry, the suggestions were applied directly in the MR even if I had not run the `make quizzes` by the moment you merged it.

> Once the MR is merged in the private repo, just run `make quizzes` on the public repo (with updated repo) and push to `main` directly without doing a PR.

Ill surely keep that in mind for the next time, it will certainly avoid possible confusions. Sorry!

python_scripts/linear_models_sol_05.py
```diff
@@ -1,11 +1,12 @@
...
+# value of `C`. In this exercise, we ask you to train a logistic regression
+# classifier using the default `penalty="l2"` regularization to find by yourself
+# the effect of the parameter `C`.
```
ogrisel, line 9:
```suggestion
# In the previous notebook we set `penalty="none"` to disable regularization
# entirely. This parameter can also control the **type** of regularization to use,
# whereas the regularization **strength** is set using the parameter `C`.
# Setting`penalty="none"` is equivalent to an infinitely large value of `C`.
# In this exercise, we ask you to train a logistic regression classifier using the
# `penalty="l2"` regularization (which happens to be the default in scikit-learn)
# to find by yourself the effect of the parameter `C`.
```

python_scripts/linear_models_sol_05.py
ogrisel: There was a misunderstanding, see my review below and my proposal to make the phrasing more explicit.

python_scripts/parameter_tuning_randomized_search.py
```diff
@@ -119,8 +119,8 @@
 # We will optimize 3 other parameters in addition to the ones we
 # optimized in the notebook presenting the `GridSearchCV`:
 #
-# * `l2_regularization`: it corresponds to the constant to regularized the loss
-#   function
+# * `l2_regularization`: it corresponds to the constant that regularizes the loss
+#   function;
```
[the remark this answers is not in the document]
ArturoAmorQ: I think that we may have not yet mentioned ridge regression at this point in the course, and eve if the logistic regression has been used, the discussion about `C` comes later in the MOOC. Should we still mention them here?

python_scripts/linear_models_sol_05.py
```diff
@@ -47,7 +48,7 @@
 from sklearn.linear_model import LogisticRegression
 
 logistic_regression = make_pipeline(
-    StandardScaler(), LogisticRegression(penalty="l2"))
+    StandardScaler(), LogisticRegression(penalty="none"))
```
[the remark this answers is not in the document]
ArturoAmorQ: Ups! Copy-paste error, nice catch!

ogrisel (approved):

PR #534, opened 2022-01-12
Update available videos

ArturoAmorQ:
Two more videos were recently added to YouTube, therefore this PR updates #497.


ogrisel (approved):
Very good.

Issue #530, opened 2022-01-11
Sprinkle the MOOC with comments on what the evaluation means


GaelVaroquaux:
As mentioned by @ogrisel : we could add a small sentence in many places in the MOOC where we do model evaluation commenting on what the evaluation means: the choice of evaluation method, the unit, ... going back the real life.

ogrisel:
I put the MOOC 2.0 milestone in case someone has the time to incremental review and improve some notebooks. But closing this issue completely will probably be rescheduled to MOOC 3.0.

lesteve:
Moving this to v3.

PR #550, opened 2022-01-24
Sprinkle M1 with comments on what the evaluation means

ArturoAmorQ:
Partially addresses #530.

This PR adds comments on the `score` method in the M1: 
- the first time the method is used
- the first time we use a test set for scoring
- in the pipeline video, where care has to be taken on interpreting the result naively

### DIFF
--- a/python_scripts/02_numerical_pipeline_hands_on.py
+++ b/python_scripts/02_numerical_pipeline_hands_on.py
@@ -218,8 +218,11 @@ accuracy = model.score(data_test, target_test)
 print(f"Accuracy of logistic regression: {accuracy:.3f}")
 
 # %% [markdown]
-# Now the real question is: is this generalization performance relevant of a good
-# predictive model? Find out by solving the next exercise!
+# When calling the `score` method on a classification problem, the output by
+# default is the fraction of correctly classified samples. In this case, around
+# 8 / 10 of the times, the logistic regression predicts the right income of a
+# person. Now the real question is: is this generalization performance relevant
+# of a good predictive model? Find out by solving the next exercise!
 #
 # In this notebook, we learned to:
 #
--- a/python_scripts/02_numerical_pipeline_hands_on.py
+++ b/python_scripts/02_numerical_pipeline_hands_on.py
@@ -218,8 +218,8 @@ accuracy = model.score(data_test, target_test)
 print(f"Accuracy of logistic regression: {accuracy:.3f}")
 
 # %% [markdown]
-# When calling the `score` method on a classification problem, the output by
-# default is the fraction of correctly classified samples. In this case, around
+# In scikit-learn, the `score` method of a classification model returns the accuracy,
+# i.e. the fraction of correctly classified samples. In this case, around
 # 8 / 10 of the times, the logistic regression predicts the right income of a
 # person. Now the real question is: is this generalization performance relevant
 # of a good predictive model? Find out by solving the next exercise!
--- a/python_scripts/02_numerical_pipeline_introduction.py
+++ b/python_scripts/02_numerical_pipeline_introduction.py
@@ -173,8 +173,8 @@ print(f"Number of correct prediction: "
 
 # %% [markdown]
 # This result means that the model makes a correct _prediction_ for
-# approximately 82 samples out of 100. But, can a model _predict_ something
-# that it already saw? In other words, can this evaluation be trusted, or is it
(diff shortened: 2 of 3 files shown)


python_scripts/02_numerical_pipeline_introduction.py
```diff
@@ -172,7 +172,10 @@
 (target == target_predicted).mean()
 
 # %% [markdown]
-# But, can this evaluation be trusted, or is it too good to be true?
+# This result means that the model makes a correct _prediction_ for
+# approximately every 82 / 100 samples. But, can a model _predict_ something
```
ogrisel, line 176:
```suggestion
# approximately 82 samples out of 100. But, can a model _predict_ something
```

python_scripts/video_pipeline.py
```diff
@@ -30,7 +30,8 @@
 data
 
 # %% [markdown]
-# We can cherry-pick some features and only retain this subset of data
+# For the sake of simplicity, we can cherry-pick some features and only retain
+# this subset of data
```
ogrisel, line 34:
```suggestion
# this subset of data:
```

python_scripts/video_pipeline.py
```diff
@@ -96,3 +97,13 @@
...
       f"{scores.mean():.3f} +/- {scores.std():.3f}")
+
+# %% [markdown]
+# ```{note}
+# In this case, around 86% of the times the pipeline correctly predicts whether
+# if the price of a house is above or below the 200_000 dollars threshold. But
+# be aware that this score was obtained by picking some features by hand, which
+# is in general a bad practice. As we will see later in the course, a good score
+# does not mean that the model is immune to other drawbacks such as the bias we
+# are probably introducing by ignoring the rest of the features.
```
ogrisel, line 108:
```suggestion
# In this case, around 86% of the times the pipeline correctly predicts whether
# the price of a house is above or below the 200_000 dollars threshold. But
# be aware that this score was obtained by picking some features by hand, which
# is not necessarily the best thing we can do for this classification task. In this
# example we can hope that fitting a complex machine learning pipelines on a
# richer set of features can improve upon this performance level.
#
# Reducing a price estimation problem to a binary classification problem with a
# single threshold at 200_000 dollars is probably too coarse to be useful in
# in practice. Treating this problem as a regression problem is probably a better
# idea. We will see later in this MOOC how to train and evaluate the performance
# of various regression models.
```

ArturoAmorQ:
Thanks for the comments @ogrisel !

python_scripts/02_numerical_pipeline_hands_on.py
```diff
@@ -218,8 +218,11 @@
 print(f"Accuracy of logistic regression: {accuracy:.3f}")
 
 # %% [markdown]
-# Now the real question is: is this generalization performance relevant of a good
-# predictive model? Find out by solving the next exercise!
+# When calling the `score` method on a classification problem, the output by
+# default is the fraction of correctly classified samples. In this case, around
```
lesteve, line 222: I think it is weird to say "when calling the `score` method ... the output by default", there is no way to change what `score` returns:

Something like this maybe (the wording can probably be improved):
```suggestion
# In scikit-learn, the `score` method of a classification model returns the accuracy,
# i.e. the fraction of correctly classified samples. In this case, around
```

python_scripts/02_numerical_pipeline_introduction.py
```diff
@@ -172,7 +172,10 @@
 (target == target_predicted).mean()
 
 # %% [markdown]
-# But, can this evaluation be trusted, or is it too good to be true?
+# This result means that the model makes a correct _prediction_ for
+# approximately 82 samples out of 100. But, can a model _predict_ something
+# that it already saw? In other words, can this evaluation be trusted, or is it
```
lesteve, line 177: I think I know what you are trying to say (that we are measuring the accuracy on the training data and that it is kind of cheating) but I find the wording super confusing ... 

In particular: "can a model predict something that it already saw?" I would answer "yes why not, sorry is this a trick question?"

I think you probably mean "can this really be called prediction when we are learning and predicting from the same data" but I can't find a good wording that convinces me.

I kind of think the next section in train-test data split explains this kind of thing already so I would stay short maybe something like this:

Note that here we used the same data to learn and evaluate our model, so can this evaluation be trusted, or is it too good to be true?

lesteve:
I tried hard to refrain from tweaking the wording but I did not manage to ...

Thanks, merging this one!

PR #551, opened 2022-01-26
Fix for pandas 1.4

lesteve:
We were using:

```py
plt.plot(dataframe_single_column, array_1d)
```

This is probably edge-casy and breaks with pandas 1.4 (released a few days ago). Whether this is a regression at the interface between pandas and matploblib is debatable. We need to fix it anyway in this repo even if it eventually gets fixed in matplotlib or pandas.

I suspect this happened by coincidence, before we were using numpy arrays, and we introduced dataframes to make scikit-learn happy about having feature names at predict time. The code was working fine until it didn't ...


ArturoAmorQ (approved):
LGTM

PR #553, opened 2022-01-28
Improve motivation of OneHotEncoder drop parameters

lesteve:
I saw that first (weird multiple "write your code here" cells) in https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_ex_04.html:
image

and ended up improving the motivations of the exercise additional correction section


ArturoAmorQ (approved):
LGTM as it is, but maybe in the future we can try to find a way to convert the categorical section into part of the exercise.

ogrisel (approved):

PR #554, opened 2022-01-28
Simplify table with confusing score/error/rank_score

lesteve:
https://mooc-forums.inria.fr/moocsl/t/error-in-hyperparameter-tuning-lecture/4530

I named the column `mean_test_error` and only kept the mean_test_error in the table

### DIFF
--- a/python_scripts/ensemble_hyperparameters.py
+++ b/python_scripts/ensemble_hyperparameters.py
@@ -59,10 +59,10 @@ search_cv = RandomizedSearchCV(
 search_cv.fit(data_train, target_train)
 
 columns = [f"param_{name}" for name in param_distributions.keys()]
-columns += ["mean_test_score", "rank_test_score"]
+columns += ["mean_test_error"]
 cv_results = pd.DataFrame(search_cv.cv_results_)
-cv_results["mean_test_score"] = -cv_results["mean_test_score"]
-cv_results[columns].sort_values(by="rank_test_score")
+cv_results["mean_test_error"] = -cv_results["mean_test_score"]
+cv_results[columns].sort_values(by="mean_test_error")
 
 # %% [markdown]
 # We can observe in our search that we are required to have a large


ArturoAmorQ:
I guess the explicit ranking column was indeed unnecessary, can we do the same for the dataframe in lines 129 to 133?

lesteve:
Indeed good catch, we seem to do something similar in python_scripts/ensemble_sol_01.py as well:
```
❯ git grep -P 'mean_test_score.+= -cv.+mean_test_score' python_scripts/
python_scripts/ensemble_hyperparameters.py:cv_results["mean_test_score"] = -cv_results["mean_test_score"]
python_scripts/ensemble_hyperparameters.py:cv_results["mean_test_score"] = -cv_results["mean_test_score"]
python_scripts/ensemble_sol_01.py:cv_results["mean_test_score"] = -cv_results["mean_test_score"]
```

ArturoAmorQ (approved):
LGTM, thanks @lesteve !

python_scripts/ensemble_hyperparameters.py
```diff
@@ -127,10 +128,11 @@
...
 
 columns = [f"param_{name}" for name in param_distributions.keys()]
-columns += ["mean_test_score", "rank_test_score"]
+columns += ["mean_test_error", "std_test_error"]
 cv_results = pd.DataFrame(search_cv.cv_results_)
-cv_results["mean_test_score"] = -cv_results["mean_test_score"]
-cv_results[columns].sort_values(by="rank_test_score")
+cv_results["mean_test_error"] = -cv_results["mean_test_score"]
+cv_results["std_test_error"] = cv_results["std_test_score"]
+cv_results[columns].sort_values(by="mean_test_error")
```
ogrisel, line 135:
```suggestion
cv_results[columns].sort_values(by="mean_test_error", ascending=False)
```

ogrisel (approved):
Ok with the proposed change but we need to restore the original ordering:

lesteve:
> Ok with the proposed change but we need to restore the original ordering:

:thinking: this is the original ordering, since rank_score is actually best score first so smaller error first.

This PR:
image

main:
image

lesteve:
Merging this one!

Issue #539, opened 2022-01-18
Review remaining posts in the forum for session 2


lesteve:
Probably in order of rough priority (**edit:** I made a pass "closed" quite a few of them and moved all the remaining ones in `priority-mooc-v2`):
- ~https://mooc-forums.inria.fr/moocsl/tag/priority-todo~
- https://mooc-forums.inria.fr/moocsl/tag/priority-mooc-v2
- ~https://mooc-forums.inria.fr/moocsl/tag/priority-nice-to-hav~

I tried to add `repo-action-done` to the posts or add a message as whisper to mention the PR or MR tracking or fixing it

There is also the possibility of using the tag `priority-mooc-v3` if we think this is not such a high priority and can be tackled in session 3.

lesteve:
I have reviewed `priority-mooc-v2` tagged posts, and all of them have been tackled or are tracked in github issues, so closing this one.

PR #557, opened 2022-01-31
Explain density argument in plt.hist

ArturoAmorQ:
Partially addresses #539.

This PR removes the `density=True` argument when comparing baselines in M7 as suggested in [this forum comment](https://mooc-forums.inria.fr/moocsl/t/density-argument-in-plt-hist/4191/2).

It also introduces a brief explanation the first time this argument appears.

Side effect: removes an unnecessary (blank) solution cell.

### DIFF
--- a/python_scripts/cross_validation_train_test.py
+++ b/python_scripts/cross_validation_train_test.py
@@ -253,15 +253,11 @@ len(cv_results)
 # %%
 import matplotlib.pyplot as plt
 
-cv_results["test_error"].plot.hist(bins=10, edgecolor="black", density=True)
+cv_results["test_error"].plot.hist(bins=10, edgecolor="black")
 plt.xlabel("Mean absolute error (k$)")
 _ = plt.title("Test error distribution")
 
 # %% [markdown]
-# The argument `density=True` in the plot above normalizes the histogram such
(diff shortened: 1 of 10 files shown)


python_scripts/cross_validation_train_test.py
```diff
@@ -258,6 +258,10 @@
 _ = plt.title("Test error distribution")
 
 # %% [markdown]
+# The argument `density=True` in the plot above normalizes the histogram such
```
lesteve, line 261: To be honest I don't think `density=True` is super useful either, but maybe that's just me ... it would be fine to show counts instead.

IMO `density=True` can lead to confusing y values since it is sensitive to the bin width.

ArturoAmorQ: I was also thinking about avoiding this everywhere in the MOOC, mostly because we never really give a probabilistic interpretation to areas.

ArturoAmorQ: I removed the `density=True` argument in all the histogram plots, what do you think?

lesteve: I am fine with this

lesteve (approved):

ogrisel (approved):

PR #562, opened 2022-02-07
Improve description on how to access help inside notebook

ArturoAmorQ:
In the [Solution for Exercise M1.02](https://inria.github.io/scikit-learn-mooc/python_scripts/02_numerical_pipeline_sol_00.html) we present how to access the description of a model using, for instance, the syntax `KNeighborsClassifier?` in this particular case.
I think the description could be improved, as it may not be clear for beginners that the `?` character was not a typo, but part of code we expected them to try.

### DIFF
--- a/python_scripts/02_numerical_pipeline_ex_00.py
+++ b/python_scripts/02_numerical_pipeline_ex_00.py
@@ -39,8 +39,13 @@ target = adult_census["class"]
 # data point.
 #
 # What is the default value of the `n_neighbors` parameter? Hint: Look at the
-# help inside your notebook `KNeighborsClassifier?` or on the [scikit-learn
+# documentation on the [scikit-learn
 # website](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html)
+# or directly access the description inside your notebook by running
+# `KNeighborsClassifier?` after the respective import.
+
+# %%
+# Write your code here.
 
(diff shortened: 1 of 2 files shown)


lesteve:
One of the problem is that the rendering is missing the pager output: 
image

Inside Jupyter notebook this will use a pager but the pager is very likely not captured by JupyterBook.

- maybe the simplest thing to do is to add a comment mentioning that `?` is intentional and will show the help. 
- is there no other place in the MOOC e.g. a youtube video that explains this kind of things?

ArturoAmorQ:
It is not captured by the JupyterBook, indeed. But I guess that is not a big problem if we alternatively show an explanation from the beginning similar to:

> What is the default value of the `n_neighbors` parameter? Hint: Look at the documentation on the [scikit-learn website](https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html) or directly access the description inside your notebook by running the following cell

```python
from sklearn.neighbors import KNeighborsClassifier

KNeighborsClassifier?
```

> This will open a pager pointing to the documentation. 

i.e., not as a #solution cell.  What do you think? Worst case scenario we can revert it after the end of the Mooc to keep a well-rendered JupyterBook.

lesteve:
> i.e., not as a #solution cell. What do you think?

:+1:

lesteve:
Merging, thanks!

PR #563, opened 2022-02-08
Fix content_only js

### LINKED ISSUE #559: js removing hypothesis in content_only mode not working
The js in `jupyter-book/_static/mooc_sklearn.js` does not seem to remove hypothesis elements anymore somehow (either when `?content_only` or inside a iframe as in FUN)

This exercise solution https://lms.fun-mooc.fr/courses/course-v1:inria+41026+session02/courseware/f59f9342fd22439fb021fd580afd962c/5d3a6873456c4a189fceac04b7a12d4c/ shows the hypothesis sidebar 

image

lesteve:
Fix #559

Turns out the problem `document.querySelector('.prev-next-bottom')` was null so this line was failing 
```
document.querySelector('.prev-next-bottom').remove();
```
and the rest of the lines were not run ... including the lines that add some CSS to hide the hypothesis sidebar.

I also ran js-beautify on the file since I was fighting against my editor and writing js is already a pain enough

### DIFF
--- a/jupyter-book/_static/sklearn_mooc.js
+++ b/jupyter-book/_static/sklearn_mooc.js
@@ -1,36 +1,43 @@
-(function(){
-  function inIframe() {
-    try {
-      return window.self !== window.top;
-    } catch (e) {
-      return true;
+(function() {
+    function inIframe() {
+        try {
+            return window.self !== window.top;
+        } catch (e) {
+            return true;
+        }
     }
-  }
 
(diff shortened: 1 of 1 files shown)


lesteve:
This seems to fix it: https://2800-246063957-gh.circle-artifacts.com/0/jupyter-book/index.html?content_only does not show any hypothesis sidebar anymore.

PR #570, opened 2022-02-13
Fix the binder launch button

### LINKED ISSUE #569: Fix a "run on binder" link on all notebook (more visible)?
This button is super useful to facilitate reusing bits of the MOOC in courses.

Currently it doesn't work (which should be easy to fix). Also, I find it hard to notice, but maybe that's only me.

GaelVaroquaux:
Closes #569


GaelVaroquaux:
OK, this seems to be working, Merging.

PR #573, opened 2022-02-14
Remove confusing lines

### LINKED ISSUE #571: set_config without import on python_scripts/02_numerical_pipeline_introduction.html
On python_scripts/02_numerical_pipeline_introduction.html there is a comment about "set_config", but because it does not give the imports, it confuses students.

GaelVaroquaux:
Fixes #571

I just removed the confusing lines, as they did not seem crucial to the message of the corresponding notebook

### DIFF
--- a/python_scripts/02_numerical_pipeline_introduction.py
+++ b/python_scripts/02_numerical_pipeline_introduction.py
@@ -92,11 +92,6 @@ print(f"The dataset contains {data.shape[0]} samples and "
 # The `fit` method is called to train the model from the input (features) and
 # target data.
 
-# %%/
-# to display nice model diagram
-from sklearn import set_config
-set_config(display='diagram')
-
 # %%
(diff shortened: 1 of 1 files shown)


glemaitre:
Actually we have the same comment that is misleading in some other place.
We should be more explicit that the diagram will be shown when calling `model` in a cell.

GaelVaroquaux:
Right, but here what is confusing is that those lines appear commented in
a strange way. Students tried copy-pasting them, but because of the
formatting did a copy-paste error, and were confused.

glemaitre:
Yes, I agree and I am merging. What I mean is that we need to find and replace the other time that we have this weird comment.

GaelVaroquaux:
> I mean is that we need to find and replace the other time that we have this weird comment.

Yes, I agree. This comment confuses the user. Maybe there is a better way
to use it.

PR #576, opened 2022-02-15
Fix notebooks not having a launch icon in JupyterBook

### LINKED ISSUE #572: Some pages do not have the "run icon" (rocket)
On the following page:
https://inria.github.io/scikit-learn-mooc/python_scripts/02_numerical_pipeline_hands_on.html
the rocket is missing. I don't know why. When teaching with binder, it is an impediment.

lesteve:
Fix #572 

I don't think this has any effects on the running MOOC since it only changes notebook metadata.

See below for the quick and dirty script I used to use the same minimal header in all the `python_scripts` files:

```py
# %%

import re
from pathlib import Path

paths = list(Path('~/dev/scikit-learn-mooc/python_scripts').expanduser().glob('*.py'))

# %%

def get_existing_header_content(content):
    header_pattern = '# ---+.+?# ---+'
    header_match = re.search(header_pattern, content, flags=re.DOTALL | re.MULTILINE)

    if not header_match:
        return ''
    
    header_content = content[header_match.start(): header_match.end()]
    return header_content

# path = Path('~/dev/scikit-learn-mooc/python_scripts/02_numerical_pipeline_sol_01.py').expanduser()
# get_existing_header_content(path.read_text())
# %%

new_header_content = """# ---
# jupyter:
#   kernelspec:
#     display_name: Python 3
#     name: python3
# ---"""

for p in paths:
    print(p)
    content = p.read_text()
    existing_header_content = get_existing_header_content(content)

    if not existing_header_content:
        new_content = new_header_content + "\n\n" + content
    else:
        new_content = content.replace(existing_header_content, new_header_content)

    p.write_text(new_content)
```

### DIFF
--- a/notebooks/01_tabular_data_exploration.ipynb
+++ b/notebooks/01_tabular_data_exploration.ipynb
@@ -483,9 +483,11 @@
   }
  ],
  "metadata": {
+  "jupytext": {
+   "main_language": "python"
+  },
   "kernelspec": {
    "display_name": "Python 3",
-   "language": "python",
    "name": "python3"
   }
  },
--- a/notebooks/01_tabular_data_exploration_ex_01.ipynb
+++ b/notebooks/01_tabular_data_exploration_ex_01.ipynb
@@ -116,11 +116,11 @@
  ],
  "metadata": {
   "jupytext": {
-   "encoding": "# -*- coding: utf-8 -*-"
+   "encoding": "# -*- coding: utf-8 -*-",
+   "main_language": "python"
   },
   "kernelspec": {
    "display_name": "Python 3",
-   "language": "python",
    "name": "python3"
   }
  },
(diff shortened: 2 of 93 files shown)


lesteve:
I checked with a local JupyterBook build that all the `python_scripts` file had the rocket launch icon:
```
 ❯ ag -L fa-rocket jupyter-book/_build/html/python_scripts       
jupyter-book/_build/html/python_scripts/helpers/__init__.html
jupyter-book/_build/html/python_scripts/helpers/plotting.html
```

so only the helper `.py` files are shown which is fine (probably we should not build html for them but that's another story)

GaelVaroquaux:
Thanks a lot. I confirmed manual that this solved the problem that I had.

Super cool! Merging!

PR #579, opened 2022-02-16
Remove utf-8 header

ArturoAmorQ:
Closes #578.

This header was not necessary. I am only removing the header from the *.py files

PR #580, opened 2022-02-16
Regenerate exercises with make exercises

lesteve:
Follow-up of #579.

This puts back exercises as generated by `make exercises`.

I looked a bit in the jupytext code, but there is no straightforward way to remove the jupytext header, see https://github.com/INRIA/scikit-learn-mooc/pull/408#issuecomment-1041463684.

I assume it's the same thing for the `utf-8 first line`. Note that the reason that `utf-8` appeared in the rendering was that they were after the yaml frontmatter due to #576 adding frontmatter at the top where there was none. Having them in the first line (or at least before the front matter) will not cause them to show up in the rendered HTML or notebook.

We could get rid of jupytext info in the frontmatter by hand if we really wanted to, there is an old TODO about this in https://github.com/INRIA/scikit-learn-mooc/blob/f5ed939ed3f259137830af1e1433bcf9f30ed5de/build_tools/generate-exercise-from-solution.py#L41-L43 but I feel this is not super high priority.

### DIFF
--- a/python_scripts/01_tabular_data_exploration_ex_01.py
+++ b/python_scripts/01_tabular_data_exploration_ex_01.py
@@ -1,5 +1,12 @@
+# -*- coding: utf-8 -*-
 # ---
 # jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.11.5
 #   kernelspec:
 #     display_name: Python 3
 #     name: python3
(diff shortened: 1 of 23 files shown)


lesteve:
> Having them in the first line (or at least before the front matter) will not cause them to show up in the rendered HTML or notebook.

Just to illustrate my point: https://2873-246063957-gh.circle-artifacts.com/0/jupyter-book/python_scripts/01_tabular_data_exploration_ex_01.html does not show utf-8 in the rendering although the `.py` file has it.

ogrisel (approved):
LGTM then.

PR #585, opened 2022-02-21
Move WIP sections to the end of _toc.yml

lesteve:


ArturoAmorQ (approved):
I guess this makes sense for keeping the content as similar as possible to the FUN version.

Issue #304, opened 2021-05-04
Review the use of significantly


lesteve:
https://mooc-forums.inria.fr/moocsl/t/review-the-use-of-significantly/556/2

PR #587, opened 2022-02-21
Remove interpretation of scores that depends on `random_state`

ArturoAmorQ:
Somewhat related to  #304. In the [randomized search notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_randomized_search.html) we still use a notion of "substantially better" based on comparing the difference of the mean test scores of two models w.r.t. the standard deviation of said scores.

Furthermore, in this case the difference depends on the `random_state` as mentioned in [this forum post](https://mooc-forums.inria.fr/moocsl/t/no-significance-between-the-2-best-models/7689/2). Therefore, the conclusions drawn are not true and are in contradiction with the following paragraph: 

> In fact, there might be other sets of parameters leading to similar or better generalization performances but that were not tested in the search.

This PR simply removes this misinterpretation.

### DIFF
--- a/python_scripts/parameter_tuning_randomized_search.py
+++ b/python_scripts/parameter_tuning_randomized_search.py
@@ -227,19 +227,6 @@ cv_results = cv_results.rename(shorten_param, axis=1)
 cv_results
 
 # %% [markdown]
-# The best model that we found with this search seems to have a substantially better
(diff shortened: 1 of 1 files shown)


ogrisel (approved):
LGTM

PR #588, opened 2022-02-21
Fix wrong step in polynomial_regression pipeline

ArturoAmorQ:
As explained in the added text, we have to set `include_bias=False` in the `PolynomialFeatures` step of the pipeline, as otherwise we would create a column perfectly correlated to the `intercept_` introduced by the following `LinearRegression`. 

I added a small crosscheck that this is equivalent to creating the features by hand up to numerical error. Feel free to comment on my wording!


ArturoAmorQ:
Another possible solution is to define the pipeline as follows:

```python
polynomial_regression = make_pipeline(
    PolynomialFeatures(degree=3),
    LinearRegression( fit_intercept=False),
)
```
but I think it is better to avoid changing the default settings of `LinearRegression`.

python_scripts/linear_regression_non_linear_link.py
```diff
@@ -177,13 +177,29 @@
...
 
+# %% [markdown]
+#
+# In the previous cell we had to set `include_bias=False` as otherwise we would
+# create a column perfectly correlated to the `intercept_` introduced by the
+# `LinearRegression`. We can verify that this procedure is equivalent to creating
+# the features by hand up to numerical error.
+
+# %%
+polynomial_regression[0].fit_transform(data) - data_expanded
```
ogrisel, line 195: Printing the all the element-wise differences for the features of the dataset generates a very large numpy array. We can summarize this with the l_inf norm:

```suggestion
np.abs(polynomial_regression[0].fit_transform(data) - data_expanded).max()
```

python_scripts/linear_regression_non_linear_link.py
```diff
@@ -177,13 +177,29 @@
...
 polynomial_regression.fit(data, target)
 target_predicted = polynomial_regression.predict(data)
 mse = mean_squared_error(target, target_predicted)
 
+# %% [markdown]
+#
+# In the previous cell we had to set `include_bias=False` as otherwise we would
+# create a column perfectly correlated to the `intercept_` introduced by the
+# `LinearRegression`. We can verify that this procedure is equivalent to creating
+# the features by hand up to numerical error.
```
ogrisel, line 192:
```suggestion
# `LinearRegression`. We can verify that this procedure is equivalent to creating
# the features by hand up to numerical error by computing the maximum
# of the absolute values of the differences between the features generated by
# both methods and checking that it is close to zero:
```

ogrisel (approved):
LGTM with the following suggestions:

python_scripts/linear_regression_non_linear_link.py
ArturoAmorQ: Perfect, thanks!

ogrisel (approved):
LGTM, thanks for the improvement!

PR #604, opened 2022-03-07
Use IPython 7 as a temporary work-around

lesteve:
This is to avoid https://github.com/ipython/ipython/issues/12467#issuecomment-1031486805.

This affects the MOOC since errors inside `joblib.Parallel` with `n_jobs > 1` looks like cells than runs forever without any output e.g. https://mooc-forums.inria.fr/moocsl/t/q7-taking-too-long-to-run-validation-curve-on-data/8023/8

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -9,6 +9,8 @@ dependencies:
   - plotly
   - jupytext
   - beautifulsoup4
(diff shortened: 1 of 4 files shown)


ogrisel:
Let's try this.

PR #608, opened 2022-03-13
to choose and not to chose

parmentelat:
here's a few changes that you might want to adopt in the wording

also I tried to `make` the .ipynb's
I noticed a few missing deps, that I recorded in the dev requirements file
there's still something fishy though, probably a dependency that needs pinning (see below)
I figured you guys'd be able to update the ipynb's separately if necessary

----
in any case here's what I get

```
$ pip show myst_parser
Name: myst-parser
Version: 0.17.0
$ make
... <snip>
  File "/Users/tparment/git/scikit-learn-mooc/build_tools/convert-python-script-to-notebook.py", line 99, in replace_admonition_in_cell_source
    config = MdParserConfig(renderer="docutils")
TypeError: __init__() got an unexpected keyword argument 'renderer'
make: *** [notebooks/01_tabular_data_exploration.ipynb] Error 1
```

### DIFF
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -292,7 +292,7 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 #
 # It is interesting to note that some machine learning models will work
 # similarly to what we did: they are known as decision tree models. The two
-# thresholds that we choose (27 years and 40 hours) are somewhat arbitrary, i.e.
+# thresholds that we chose (27 years and 40 hours) are somewhat arbitrary, i.e.
 # we choose them by only looking at the pairplot. In contrast, a decision tree
 # will choose the "best" splits based on data without human intervention or
 # inspection. Decision trees will be covered more in detail in a future module.
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -293,7 +293,7 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 # It is interesting to note that some machine learning models will work
 # similarly to what we did: they are known as decision tree models. The two
 # thresholds that we chose (27 years and 40 hours) are somewhat arbitrary, i.e.
-# we choose them by only looking at the pairplot. In contrast, a decision tree
+# we chose them by only looking at the pairplot. In contrast, a decision tree
 # will choose the "best" splits based on data without human intervention or
 # inspection. Decision trees will be covered more in detail in a future module.
 #
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -287,14 +287,14 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 #   prediction is low-income. Indeed, there are many blue points and only a few
 #   orange points.
 # * In the region `age > 27 AND hours-per-week > 40` (top-right region), we see
-#   a mix of blue points and orange points. It seems complicated to chose which
+#   a mix of blue points and orange points. It seems complicated to choose which
 #   class we should predict in this region.
 #
 # It is interesting to note that some machine learning models will work
 # similarly to what we did: they are known as decision tree models. The two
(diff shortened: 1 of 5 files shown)


python_scripts/ensemble_gradient_boosting.py
```diff
@@ -239,7 +239,7 @@ def generate_data(n_samples=50):
 print(f"Error of the tree: {target_true - y_pred_first_and_second_tree:.3f}")
 
 # %% [markdown]
-# We chose a sample for which only two trees were enough to make the perfect
+# We choose a sample for which only two trees were enough to make the perfect
```
ArturoAmorQ, line 242: I think this one should actually be in past tense, as it refers to the sample we chose above.

python_scripts/feature_selection_introduction.py
```diff
@@ -30,7 +30,7 @@
 )
 
 # %% [markdown]
-# We chose to create a dataset with two informative features among a hundred.
+# We choose to create a dataset with two informative features among a hundred.
 # To simplify our example, we did not include either redundant or repeated
```
ArturoAmorQ, line 34: To be consistent with your modification, we have to change this line as well.
```suggestion
# To simplify our example, we do not include either redundant or repeated
```

python_scripts/linear_models_regularization.py
```diff
@@ -287,7 +287,7 @@
 # ```
 #
 # In the previous analysis, we did not study if the parameter `alpha` will have
-# an effect on the performance. We chose the parameter beforehand and fix it
+# an effect on the performance. We choose the parameter beforehand and fix it
```
ArturoAmorQ, line 290:
```suggestion
# an effect on the performance. We chose the parameter beforehand and fixed it
```

python_scripts/01_tabular_data_exploration.py
```diff
@@ -287,14 +287,14 @@
...
+#   a mix of blue points and orange points. It seems complicated to choose which
 #   class we should predict in this region.
 #
 # It is interesting to note that some machine learning models will work
 # similarly to what we did: they are known as decision tree models. The two
-# thresholds that we chose (27 years and 40 hours) are somewhat arbitrary, i.e.
-# we chose them by only looking at the pairplot. In contrast, a decision tree
-# will chose the "best" splits based on data without human intervention or
+# thresholds that we choose (27 years and 40 hours) are somewhat arbitrary, i.e.
+# we choose them by only looking at the pairplot. In contrast, a decision tree
```
ArturoAmorQ, line 296: I think this two lines refer to what was done above and should not be modified.

ArturoAmorQ (commented):
Some small comments, otherwise LGTM, thanks @parmentelat

ArturoAmorQ:
And yes, we will update the notebooks later, don;t worry about that.

python_scripts/ensemble_gradient_boosting.py
parmentelat: fair enough

python_scripts/linear_models_regularization.py
ArturoAmorQ: This line should be in past tense, as we are referring to "the previous analysis"

python_scripts/01_tabular_data_exploration.py
```diff
@@ -287,14 +287,14 @@
...
-#   a mix of blue points and orange points. It seems complicated to chose which
+#   a mix of blue points and orange points. It seems complicated to choose which
 #   class we should predict in this region.
 #
 # It is interesting to note that some machine learning models will work
 # similarly to what we did: they are known as decision tree models. The two
 # thresholds that we chose (27 years and 40 hours) are somewhat arbitrary, i.e.
-# we chose them by only looking at the pairplot. In contrast, a decision tree
-# will chose the "best" splits based on data without human intervention or
+# we choose them by only looking at the pairplot. In contrast, a decision tree
```
ArturoAmorQ, line 296:
```suggestion
# we chose them by only looking at the pairplot. In contrast, a decision tree
```

parmentelat:
if anything is still not ok with you please just merge and tweak it as you see fit

PR #609, opened 2022-03-14
Fix learning curve to show overlapped error bars

ArturoAmorQ:
As mentioned in [this forum post](https://mooc-forums.inria.fr/moocsl/t/suggestion-in-exercice-m2-01-add-transparency-to-plt-errorbar/8969):

> people could believe than there are no standard deviations for training scores for gammas <1 because on the graph these std are hidden by the testing scores’ ones. Adding transparency to the graph with the alpha parameter fixes it.

### DIFF
--- a/python_scripts/cross_validation_sol_01.py
+++ b/python_scripts/cross_validation_sol_01.py
@@ -121,9 +121,9 @@ train_scores, test_scores = validation_curve(
 import matplotlib.pyplot as plt
 
 plt.errorbar(gammas, train_scores.mean(axis=1),
-             yerr=train_scores.std(axis=1), label='Training score')
+             yerr=train_scores.std(axis=1), alpha = 0.95, label='Training score')
 plt.errorbar(gammas, test_scores.mean(axis=1),
-             yerr=test_scores.std(axis=1), label='Testing score')
+             yerr=test_scores.std(axis=1), alpha = 0.5, label='Testing score')
 plt.legend()
 
 plt.xscale("log")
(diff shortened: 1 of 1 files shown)


python_scripts/cross_validation_sol_01.py
```diff
@@ -155,18 +164,27 @@
...
 _ = plt.title("Learning curve for support vector machine")
 
 # %% [markdown] tags=["solution"]
-# We observe that adding new samples in the dataset does not improve the
-# testing score. We can only conclude that the standard deviation of
-# the training error is decreasing when adding more samples which is not a
-# surprise.
+# We observe that adding new samples in the dataset does not improve the testing
+# score. We can only conclude that the standard deviation of the training error
+# is decreasing when adding more samples which is not a surprise.
```
ogrisel, line 190: Actually at this point in the MOOC, it's really not trivial to judge whether this should be surprising or not.

Instead let's conclude with something like:

```suggestion
# We observe that adding new samples to the training dataset does
# not seem to improve the training and testing scores.
#
# In particular, 76% accuracy is the score of a model that
# always predicts the majority class `"not donated"`. This can
# mean that our small pipeline is not able to use input features
# to improve upon that simplistic baseline, and increasing the
# training set size does not help either.
#
# It could be the case that the input features are fundamentally
# not very informative and the classification problem is
# fundamentally impossible to solve to a high accuracy. But it
# could also be the case that our choice of using the default
# hyperparameter value of the `SVC` class was a bad idea, or that
# the choice of the `SVC` class is itself sub-optimal.
#
# Later in this MOOC we will see how to better tune the
# hyper-parameters of a model and explore how to compare the
# predictive performance of different model classes in a more
# systematic way.
```

ogrisel (approved):
+1 for the opacity fix.

Let me suggest a further improvement to the concluding paragraph in the solution:

lesteve:
FYI I updated the exercise with `make exercises` and all the notebooks in 7b4d2cb7

PR #610, opened 2022-03-17
Remove not super useful set_config diagram

lesteve:
https://inria.github.io/scikit-learn-mooc/python_scripts/02_numerical_pipeline_hands_on.html

The only place this is used is here:
image

I think it is fine to start using HTML diagrams in the next notebook where at least there is a pipeline.

If we want to keep it, we may put the `set_config` code somewhere else (e.g. at the beginning of the notebook) since with the sentence before it looks like this is part of creating the `LogisticRegression`


ArturoAmorQ:
I agree, thanks @lesteve!

PR #615, opened 2022-03-31
Text does not match regression tree predictions

ArturoAmorQ:
As mentioned in [this forum post](https://mooc-forums.inria.fr/moocsl/t/contradiction-between-text-and-tree-structure-diagram/9513), we have a mismatch between text and the diagram in the [Decision trees for regression notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_regression.html). 

In the tree structure diagram:
left leaf node: **value = 3698.709**
right leaf node: **value = 5032.364**

In text they are reported as:
The predicted values on each side of the split are two constants: **3683.50 g** and **5023.62 g**

This PR fixes it and, as always, makes unrelated minor formatting.


lesteve:
Merging, thanks!

PR #617, opened 2022-04-01
Improve wording in definition of bootstrap resampling

ArturoAmorQ:
As discussed in [this forum post](https://lms.fun-mooc.fr/courses/course-v1:inria+41026+session02/6ad122825f0f4241a47f1511411e65c8/), the wording we use for the definitions of bootstrapping and bootstrap samples can be clearer.

Here is a first approach of a better wording, but feedback is always welcome.


python_scripts/ensemble_bagging.py
```diff
@@ -83,10 +83,16 @@ def generate_data(n_samples=30):
...
 # ## Bootstrap resampling
 #
-# A bootstrap sample corresponds to a resampling with replacement, of the
-# original dataset, a sample that is the same size as the original dataset.
-# Thus, the bootstrap sample will contain some data points several times while
-# some of the original data points will not be present.
+# Bootstrapping is a resampling "with replacement" of the original
+# dataset. It corresponds to sampling n out of n data points with
+# replacement uniformly at random from the original dataset. n is the
+# number of data points in the original dataset.
```
ArturoAmorQ, line 89:
```suggestion
# Given a dataset with `n` data points, bootstrapping corresponds
# to resampling "with replacement"  `n` out of such `n` data points
# uniformly at random.
```

PR #622, opened 2022-04-25
Update yml file in local install instructions

ArturoAmorQ:
As mentioned in [this forum post](https://mooc-forums.inria.fr/moocsl/t/get-feature-names-out/10331) our local install instruction led people to use `sklearn version 0.24.2`, which has some deprecated notations.

This PR fixes it.


local-install-instructions.md
```diff
@@ -46,7 +46,7 @@ Using python in /home/lesteve/miniconda3/envs/scikit-learn-course
 [ OK ] numpy version 1.19.5
 [ OK ] scipy version 1.6.0
 [ OK ] matplotlib version 3.3.3
-[ OK ] sklearn version 0.24.0
+[ OK ] sklearn version 1.1.2
```
glemaitre, line 49:
```suggestion
[ OK ] sklearn version 1.0.2
```

lesteve:
The link to the forum post seems off: it goes to https://lms.fun-mooc.fr/courses/course-v1:inria+41026+session02/6ad122825f0f4241a47f1511411e65c8/ ... can you correct it?

If we indeed require scikit-learn >= 1 for some of our notebooks, we should also update https://github.com/INRIA/scikit-learn-mooc/blob/main/check_env.py

ArturoAmorQ:
> The link to the forum post seems off: it goes to https://lms.fun-mooc.fr/courses/course-v1:inria+41026+session02/6ad122825f0f4241a47f1511411e65c8/ ... can you correct it?
> 
> If we indeed require scikit-learn >= 1 for some of our notebooks, we should also update https://github.com/INRIA/scikit-learn-mooc/blob/main/check_env.py

You are right. It's done, thanks!

lesteve:
You may as well do something similar in `requirements*.txt` and `environment-dev.yml`. I would use `1.0.2` everywhere for consistency.

lesteve:
Merging, thanks!

Issue #427, opened 2021-08-03
Add hyperlinks for clarity whenever possible


ArturoAmorQ:
Add hyperlinks whenever a new concept is introduced (and not covered in detail), or as a reminder in key places. 
This may help clarifying the content without adding more text. 

- Most of the times the concepts are covered in the [scikit-learn user guide](https://scikit-learn.org/stable/user_guide.html).
Example:
    - Add link to the documentation for terms [mean squared error](https://scikit-learn.org/stable/modules/model_evaluation.html#mean-squared-error), [mean absolute error](https://scikit-learn.org/stable/modules/model_evaluation.html#mean-absolute-error) and [R^2](https://scikit-learn.org/stable/modules/model_evaluation.html#r2-score-the-coefficient-of-determination) in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_regression_in_sklearn.html)
    - For the term [API](https://scikit-learn.org/stable/glossary.html#term-API) introduced at the beginning of M1.
    
- Some concepts are not in the [user guide](https://scikit-learn.org/stable/user_guide.html) but we could find a way to link them to the [Glossary](https://inria.github.io/scikit-learn-mooc/appendix/glossary.html).
Examples:
    - Add link to the [Glossary](https://inria.github.io/scikit-learn-mooc/appendix/glossary.html) for the word "uncertainty" in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/02_numerical_pipeline_scaling.html) (See #426)
    - Add link to the [Glossary](https://inria.github.io/scikit-learn-mooc/appendix/glossary.html) the first time we introduce the word "hyperparameter" in [this notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_learning_curve.html)  and in quiz [M3.01](https://github.com/INRIA/scikit-learn-mooc/blob/master/jupyter-book/tuning/parameter_tuning_manual_quiz_m3_01.md) (to address [this post in the forum](https://mooc-forums.inria.fr/moocsl/t/hyperparameters/2470/4))
 
- Other concepts may be pointed to external material "to go beyond", e.g. ["interpretability"](https://mooc-forums.inria.fr/moocsl/t/had-to-look-up-the-definition-of-interpretability-for-this-one/10524) or ["voting"](https://mooc-forums.inria.fr/moocsl/t/make-them-vote/11012).

lesteve:
[...]

PR #623, opened 2022-04-25
Fix typo s/might have reach/might have reached

kinow:
Hi,

Quick fix for a typo. Also, I had to Google [Bayes error rate](https://en.wikipedia.org/wiki/Bayes_error_rate) after reading this paragraph, but I couldn't find a link in the notebook for that. Are there any plans to have footnotes (like in GitHub Markdown [^1]) with useful links? Or maybe these links are in the forum or somewhere else?

Thanks!

[^1]: for example

### DIFF
--- a/python_scripts/cross_validation_learning_curve.py
+++ b/python_scripts/cross_validation_learning_curve.py
@@ -111,7 +111,7 @@ _ = plt.title("Learning curve for decision tree")
 # more samples into the training set.
 #
 # If we achieve a plateau and adding new samples in the training set does not
-# reduce the testing error, we might have reach the Bayes error rate using the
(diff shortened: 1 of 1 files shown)


ArturoAmorQ:
> Are there any plans to have footnotes (like in GitHub Markdown [1](https://github.com/INRIA/scikit-learn-mooc/pull/623#user-content-fn-1-d9f3425293c658d5fbebdbe2a52b8a32)) with useful links? Or maybe these links are in the forum or somewhere else?

I think having footnotes is a good idea, maybe you can join the discussion in Issue #427 and add the term "Bayes error rate" there.

ArturoAmorQ (approved):
On the meantime, this PR LGTM as it is, thanks for the contribution @kinow!

PR #634, opened 2022-05-23
DecisionBoundaryDisplay

ArturoAmorQ:
Implements [`sklearn.inspection.DecisionBoundaryDisplay`](https://scikit-learn.org/dev/modules/generated/sklearn.inspection.DecisionBoundaryDisplay.html) instead of the workaround [`helpers.plotting`](https://github.com/INRIA/scikit-learn-mooc/blob/main/python_scripts/helpers/plotting.py).


python_scripts/logistic_regression.py
```diff
@@ -100,16 +100,12 @@
 # ```{note}
 # Here, we will use the class `DecisionBoundaryDisplay`. We provide this class
 # to allow making plots of the decision function boundary in a 2 dimensional
-# space. The implementation can be found [here](
-# https://github.com/INRIA/scikit-learn-mooc/blob/main/python_scripts/helpers/plotting.py).
-# This class is intended to be part of the `scikit-learn` package in the future
-# as it is proposed in the following [Pull-Request](
-# https://github.com/scikit-learn/scikit-learn/pull/16061).
+# space.
```
ogrisel, line 103: If this is the first time we use this tool, maybe we can give motivation for the use of this class and better manage expectations:


```suggestion
# space. This educational utility allows us to gain some insights on
# the shape of the class decision boundary learned by the classifier.
```

then at the end of this notebook one could add a paragraph such as:


```
# The class `DecisionBoundaryDisplay` makes it possible to visualize
# the shape of the decision function of a classifier trained to
# classify data points in a 2D feature space.
#
# Note however that in more realistic machine learning contexts, one
# would typically fit a more on more than two features at once and
# therefore it would not be possible to display such a visualization of
# of the decision boundary in general.
```

ogrisel: Also, the sentence:

```
# Indeed, both coefficients are non-null.
```

could be expanded a bit:

```
# Indeed, both coefficients are non-null. If one of them had been zero,
# the decision boundary would have been either horizontal or vertical.
#
# Furthermore the intercept is also non-zero, which means that the
# decision does not go through the point with (0, 0) coordinates.
#
# For the mathematically inclined reader, the equation of the decision
# boundary is:
#
#     coef0 * x0 + coef1 * x1 + intercept = 0
#
# where x0 is "Culmen Length (mm)" and x1 is "Culmen Depth (mm)". 
#
# This equation is equivalent to (assuming that coef1 is non-zero):
#
#     x1 = coef0 / coef1 * x0 - intercept / coef1
#
# which is the equation of a straight line.
```

ogrisel (approved):
I checked the rendered HTML and it looks good. Here are some small suggestions for the logistic regression case:

python_scripts/logistic_regression.py
```diff
@@ -136,4 +137,21 @@
...
+# For the mathematically inclined reader, the equation of the decision boundary
+# is:
+#
+#     `coef0 * x0 + coef1 * x1 + intercept = 0`
+#
+# where `x0` is `"Culmen Length (mm)"` and `x1` is `"Culmen Depth (mm)"`.
+#
+# This equation is equivalent to (assuming that `coef1` is non-zero):
+#
+#     `x1 = coef0 / coef1 * x0 - intercept / coef1`
```
ogrisel, line 155:
```suggestion
#     x1 = coef0 / coef1 * x0 - intercept / coef1
```

ogrisel (approved):
Based on the HTML rendering, the quotes for the equations are not needed:

https://output.circle-artifacts.com/output/job/da957598-65eb-4bd4-b267-9eddb24c68a4/artifacts/0/jupyter-book/python_scripts/logistic_regression.html

PR #637, opened 2022-05-30
Update yml file in local install instructions

ArturoAmorQ:
Updates #622.

Updates local install instructions to `sklearn version 1.1.1`.


check_env.py
```diff
@@ -55,7 +55,7 @@ def import_version(pkg, min_ver, fail_msg=""):
 
 print()
 requirements = {'numpy': "1.16", 'scipy': "1.2", 'matplotlib': "3.0",
-                'sklearn': "1.0", 'pandas': "1",
+                'sklearn': "1.1.1", 'pandas': "1",
```
ogrisel, line 58: I think we do not need to pin a specific bugfix versions. Let's automatically use the latest.

```suggestion
                'sklearn': "1.1", 'pandas': "1",
```

ogrisel (approved):
Small comment, but otherwise LGTM.

PR #638, opened 2022-06-13
Use Plus-Minus Sign Unicode Character

ArturoAmorQ:
This PR changes the notation +/- in the python scripts to use the ± U+00B1 Plus-Minus Sign Unicode Character.


ogrisel (approved):
Nice!

PR #650, opened 2022-07-11
use direct weblink to dataset

psteinb:
This way, the cells can be run from anywhere with internet connectivity.

### DIFF
--- a/notebooks/01_tabular_data_exploration.ipynb
+++ b/notebooks/01_tabular_data_exploration.ipynb
@@ -49,7 +49,7 @@
    "source": [
     "import pandas as pd\n",
     "\n",
-    "adult_census = pd.read_csv(\"../datasets/adult-census.csv\")"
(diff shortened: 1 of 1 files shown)


lesteve:
Short summary: I am not in favour of this change.
- I can see why this is better in a way since the notebook is more stand-alone and does not need the local dataset.
- this is also worse in another way because now you need internet to run the notebook.
- this is also a bit weird in the sense that you always use the latest version of the dataset from github rather than the local one which is associated to the current commit you are on. For example, for the MOOC use case changing the dataset (this does not happen very often to be fair) would affect people depending when they run the notebook, which is not really something we want ...

As indicated in the README, I think people who want to run the notebooks locally should follow https://github.com/INRIA/scikit-learn-mooc/blob/main/local-install-instructions.md.


Of course if I missed something, or if others have a different opinion, let me know!

PR #652, opened 2022-07-18
reduce the complexity of language

psteinb:
- avoid long sentences
- add more repetition of important connections
- add more references to what is meant in the notebook
- one small fix to orthography

Would be interesting to hear your review on these small changes.

### DIFF
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -28,7 +28,7 @@
 # <http://www.openml.org/d/1590>
 #
 # The dataset is available as a CSV (Comma-Separated Values) file and we will
-# use pandas to read it.
+# use `pandas` to read it.
 #
 # ```{note}
 # [Pandas](https://pandas.pydata.org/) is a Python library used for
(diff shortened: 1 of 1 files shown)


python_scripts/01_tabular_data_exploration.py
```diff
@@ -99,11 +101,22 @@
...
-    "workclass", "education", "marital-status", "occupation",
-    "relationship", "race", "sex", "native-country"]
+    "workclass",
+    "education",
+    "marital-status",
+    "occupation",
+    "relationship",
+    "race",
+    "sex",
+    "native-country",
```
GaelVaroquaux, line 118: I personnally am not super excited about those cosmetic changes.

First: this formatting makes sense maybe to reduce merge conflicts, but it also maximizes vertical scrolling. So for the reader it is not optimal, in particular for the reader that is not a trained software engineer used to this formatting.

Second, if we all start changing these details, we'll be spending a lot of time on them (possibly with back and forth according to people's preference).

GaelVaroquaux (approved):
Thanks for the changes. I'm merging.

Let's also avoid changing too many minor details such as where line returns are.

Issue #603, opened 2022-03-07
Tuning n_estimators for RandomForests doesn't make sense


GaelVaroquaux:
In the following:
https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_hyperparameters.html
We tune n_estimators for RandomForest which doesn't make much sense. Typically: the more, the better.

GaelVaroquaux:
I realize that this is somewhat mentioned below, but not very clearly IMHO. I think that it should be clearly stated that tuning the number of estimators for RandomForests does not make much sense and is likely to be a loss of resources.

ogrisel:
But we don't have a smart API to explore efficiently how many trees are needed.

We could do a big fit and then do an a posteriori analysis of how many trees are actually necessary by subsampling the `estimators_` attribute (and re-setting `n_estimators` for consistency) in a clone but this not standard scikit-learn code.

PR #658, opened 2022-09-08
Emphasize good practice of not tuning `n_estimators`

ArturoAmorQ:
Fixes [#603](https://github.com/INRIA/scikit-learn-mooc/issues/603).

This PR implements the conservative solution proposed [in this comment](https://github.com/INRIA/scikit-learn-mooc/issues/603#issuecomment-1060471226) at the expense of the more radical solution proposed in [this other comment](https://github.com/INRIA/scikit-learn-mooc/issues/603#issuecomment-1068926935).


python_scripts/ensemble_hyperparameters.py
```diff
@@ -142,12 +149,6 @@
 cv_results[columns].sort_values(by="mean_test_error")
 
 # %% [markdown]
-#
-# ```{caution}
-# Here, we tune the `n_estimators` but be aware that using early-stopping as
-# in the previous exercise will be better.
-# ```
-#
```
ogrisel, line 150: I think we should keep this note here as using early stopping for `GradientBoosting*` models is often a good idea but grid-searching `n_estimators` is an ok-ish alternative to me.

ogrisel (commented):
Some more feed back but otherwise LGTM. Thanks @ArturoAmorQ.

PR #661, opened 2022-09-15
Remove averaging over optimal `alpha` regularization in RidgeCV

### LINKED ISSUE #642: Averaging Ridge's `alpha` over the CV folds
As mentioned [here in the forum](https://mooc-forums.inria.fr/moocsl/t/averaging-alpha-over-the-cv-folds/8936/2), we should probably change content of the [linear regularization notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_regularization.html) to not recommend to average the minimum of the best alpha values on each CV fold but instead average the CV scores of each CV fold and then take the value of alpha that minimize the average test score.

ArturoAmorQ:
Fixes #642.


glemaitre (approved):
I am OK with these changes. They are indeed better than the previous version.

PR #663, opened 2022-10-07
Improve paragraph on data imbalance

ArturoAmorQ:
Previous wording was not exact.


ogrisel (approved):
LGTM :)

ogrisel:
Maybe, that we could add a sentence after the paragraph pointing to fairlearn such as:

"Studying why the data collection process of this dataset lead to such an unexpected gender imbalance is beyond the scope of this MOOC but we should keep in mind that this dataset is not representative of the US population before drawing any conclusions based on its statistics or the predictions of models trained on it."

glemaitre:
LTGM. Thanks @ArturoAmorQ

PR #667, opened 2022-10-14
Add mention to OneHotEncoder parameter `infrequent_if_exist`

ArturoAmorQ:
We could demo the use of `handle_unknown="ignore"` and `handle_unknown="infrequent_if_exist"`.

Notice that the Wrap-up quiz M1 requires setting `handle_unknown="ignore"` and, as I do not intend to introduce last minute changes to a quiz, this PR only mentions the parameter introduced in 1.1.


python_scripts/03_categorical_pipeline.py
```diff
@@ -130,63 +133,63 @@
...
+# will be set to `0`.
 #
 # We will start by encoding a single feature (e.g. `"education"`) to illustrate
 # how the encoding works.
 
 # %%
 from sklearn.preprocessing import OneHotEncoder
 
-encoder = OneHotEncoder(sparse_output=False)
+encoder = OneHotEncoder(sparse=False)
```
ogrisel, line 174: `sparse_output` was the correct parameter.

python_scripts/03_categorical_pipeline.py
```diff
@@ -276,30 +278,36 @@
...
 #
 # * list all the possible categories and provide it to the encoder via the
 #   keyword argument `categories`;
-# * use the parameter `handle_unknown`, i.e. if an unknown category is
+# * set the parameter `handle_unknown="infrequent_if_exist"`, i.e. if one or
+#   more unknown categories are encountered during transform, the resulting
+#   one-hot encoded columns will be mapped to the last position in the encoding.
+#   The notion of what is "infrequent" can be controlled with the
+#   hyper-parameter `min_frequency`.
+# * set the parameter `handle_unknown="ignore"`, i.e. if an unknown category is
```
ogrisel, line 295: I think we should put `handle_unknown="ignore"` before `handle_unknown="infrequent_if_exist"` because `"ignore"` is easier to understand.

python_scripts/03_categorical_pipeline.py
```diff
@@ -276,30 +278,36 @@
...
+# We see that the `"Holand-Netherlands"` category is occurring rarely. This will
 # be a problem during cross-validation: if the sample ends up in the test set
 # during splitting then the classifier would not have seen the category during
 # training and will not be able to encode it.
 #
-# In scikit-learn, there are two solutions to bypass this issue:
+# In scikit-learn, there are some possible solutions to bypass this issue:
 #
 # * list all the possible categories and provide it to the encoder via the
 #   keyword argument `categories`;
```
ogrisel, line 289:
```suggestion
#   keyword argument `categories` instead of letting the estimator try to
#   automatically find this list only from the training side of the split when
#   calling `fit`;
```

python_scripts/03_categorical_pipeline.py
```diff
@@ -276,30 +278,36 @@
...
-# * use the parameter `handle_unknown`, i.e. if an unknown category is
+# * set the parameter `handle_unknown="infrequent_if_exist"`, i.e. if one or
+#   more unknown categories are encountered during transform, the resulting
+#   one-hot encoded columns will be mapped to the last position in the encoding.
+#   The notion of what is "infrequent" can be controlled with the
+#   hyper-parameter `min_frequency`.
```
ogrisel, line 294: I think it would be easier to understand if we first explain what `min_frequency` does to then explain what `handle_unknown="infrequent_if_exist"` means.

```suggestion
# * it is possible to set `min_frequency` to a non-default value to
#   collapse the rarest categories observed on the training data
#   into a the last one-hot encoded feature. When this option is
#   enabled, it is then possible to also set
#   `handle_unknown="infrequent_if_exist"` to also encode the
#   unknown categories (categories only observed at predict time)
#   as ones in that last column.
```

python_scripts/03_categorical_pipeline.py
```diff
@@ -276,30 +278,37 @@
...
 # training and will not be able to encode it.
 #
-# In scikit-learn, there are two solutions to bypass this issue:
+# In scikit-learn, there are some possible solutions to bypass this issue:
 #
-# * list all the possible categories and provide it to the encoder via the
-#   keyword argument `categories`;
-# * use the parameter `handle_unknown`, i.e. if an unknown category is
+# * list all the possible categories and provide them to the encoder via the
+#   keyword argument `categories` instead of letting the estimator to
```
ogrisel, line 289: Grammar:

```suggestion
#   keyword argument `categories` instead of letting the estimator
```

python_scripts/03_categorical_pipeline.py
```diff
@@ -276,30 +278,37 @@
...
-#   feature will be all zeros.
+#   feature will be all zeros;
+# * adjust the `min_frequency` parameter to collapse the rarest categories
+#   observed in the training data into a single one-hot encoded feature. If you
+#   enable this option, you can also set `handle_unknown="infrequent_if_exist"`
+#   to encode the unknown categories (categories only observed at predict time)
+#   as ones in that last column.
 #
-# Here, we will use the latter solution for simplicity.
+# Here we demo the second option, but feel free to evaluate the alternatives.
```
ogrisel, line 300: I find that "here" followed by the note about `OrdinalEncoder` breaks makes it hard to follow the though process.

```suggestion
# In this notebook we only explore the second option,
# `OneHotEncoder(handle_unknown="ignore")`, but feel free to evaluate the
# alternatives on your own, for instance using a sandbox notebook.
```

python_scripts/03_categorical_pipeline.py
```diff
@@ -276,30 +278,37 @@
...
 # %% [markdown]
 # ```{tip}
-# Be aware the `OrdinalEncoder` exposes as well a parameter `handle_unknown`. It
-# can be set to `use_encoded_value`. If that option is chosen, you can define a
-# fixed value to which all unknowns will be set to during `transform`. For
-# example, `OrdinalEncoder(handle_unknown='use_encoded_value',
-# unknown_value=42)` will set all values encountered during `transform` to `42`
-# which are not part of the data encountered during the `fit` call. You are
-# going to use these parameters in the next exercise.
+# Be aware the `OrdinalEncoder` exposes as well a parameter
```
ogrisel, line 304:
```suggestion
# Be aware the `OrdinalEncoder` exposes a parameter also named
```

ogrisel (approved):
Some more details, otherwise, LGMT.

PR #674, opened 2022-12-05
MAINT Use class `PredictionErrorDisplay`

ArturoAmorQ:
[`PredictionErrorDisplay`](https://scikit-learn.org/dev/modules/generated/sklearn.metrics.PredictionErrorDisplay.html) is available since 1.2.0rc1. This PR makes use of this new class for easing the plotting a little bit.

~~Do not merge until scikit-learn v1.2 is implemented for the MOOC.~~

It's good to merge now.


lesteve:
I converted it to draft, it should probably not be merged until we stop the MOOC JupyterHub for the third session, to avoid weird side-effects.

I don't remember off the top of my head whether the docker image is updated or not for existing users. It may be the case, which is not such a good idea but oh well ...

python_scripts/metrics_regression.py
```diff
@@ -148,20 +148,20 @@
...
-plt.axline((0, 0), slope=1, label="Perfect fit")
-plt.axis('square')
+from sklearn.metrics import PredictionErrorDisplay
+
+PredictionErrorDisplay.from_predictions(
+    y_true=target_test,
+    y_pred=target_predicted,
+    kind="actual_vs_predicted",
+    subsample=100,
+    random_state=0,
```
ArturoAmorQ, line 160:
```suggestion
    scatter_kwargs={"alpha": 0.5},
```

ogrisel (approved):
LGTM. I think we could also add the residuals vs predicted, side to side with the actual vs predicted plot.

python_scripts/metrics_regression.py
```diff
@@ -185,21 +202,33 @@
...
+axs[1].set_xlabel("Predicted values (k$)")
+axs[1].set_ylabel("Residual values (k$)")
+
+_ = fig.suptitle("Regression using a model that\ntransforms the target "
+                 "before fitting", y=1.1)
 
 # %% [markdown]
 # Thus, once we transformed the target, we see that we corrected some of the
-# high values.
+# predictions of the high values.
```
ogrisel, line 234: The model with the transformed target seems to exhibit fewer structure in its residuals: over-estimation and under-estimation errors seems to be more balanced.

We can confirm this by computing the previous metrics for that last model and observe that they all improved w.r.t. the linear regression model without the target transformation.

TODO: add new cells to compute all the metrics for the last model.

While a common practice, performing such a target transformation for linear regression is often frowned upon by statisticians: it is often mathematically more
justified to instead adapt the loss function of the regression model itself, for
instance by fitting a [PoissonRegressor](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.PoissonRegressor.html) or a [TweedieRegressor](
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.TweedieRegressor.html) model instead of `LinearRegression`. In particular those
models indeed use an internal "log link" function that makes them more suited for
this kind of positive-only target data distributions but this analysis is beyond the scope of this MOOC.

The interested readers are encouraged to learn more about those models, in particular by reading the user guide and the examples linked for their API documentation pages above.

ogrisel (commented):
Thanks this is great. Here are more suggestions for further improvements.

python_scripts/metrics_regression.py
```diff
@@ -185,21 +206,62 @@
...
+# justified to instead adapt the loss function of the regression model itself,
+# for instance by fitting a `PoissonRegressor` or a `TweedieRegressor` model
+# instead of `LinearRegression`. In particular those models indeed use an
+# internal "log link" function that makes them more suited for this kind of
+# positive-only target data distributions, but this analysis is beyond the scope
+# of this MOOC.
+#
+# The interested readers are encouraged to learn more about those models, in
+# particular by reading the user guide and the examples linked for their API
+# documentation pages above.
```
ogrisel, line 267: Thanks for integrating my suggestion. However we need to make the class names for `PoissonRegressor` and `TweedieRegressor` linked to the scikit-learn API documentation for each class.

Or alternatively we should change this last paragraph to link to the user guide on generalized linear models.

ogrisel (commented):
Thanks @ArturoAmorQ. The new plots look great but we miss some links to the the scikit-learn doc in the second to last or the last paragraph.

python_scripts/metrics_regression.py
ArturoAmorQ: I chose the former as the user guide on generalized linear models contains some other models (including the logistic regression) and may bring more questions than answers.

python_scripts/metrics_regression.py
```diff
@@ -157,39 +158,56 @@
...
-
-# %%
-predicted_actual = {
-    "True values (k$)": target_test,
-    "Predicted values (k$)": target_predicted,
-}
-predicted_actual = pd.DataFrame(predicted_actual)
+# predicted values versus the true values. On such a plot, correct predictions
+# lie on the diagonal line. This allows us to detect if the model makes errors
+# in a consistent way, i.e. has some bias.
```
ogrisel, line 163:
```suggestion
# predicted values versus the true values.
#
# In the hypothetical case wher all the variations of `y` could be perfectly
# explained by `X` (no unobserved factor of variations) and if we had chosen
# an optimal regressor class and hyper-parameters then we would expect all
# its predictions to lie on the diagonal line of the first plot below.
#
# In the real life, this is almost never the case: some unknown fraction of the
# variations in `y` cannot be explained by variations in `X`: they stem from
# external factors not represented by the columns `X`.
#
# Therefore the best we can hope for, is that the predictions of our model
# lie in a cloud of points approximately symmetrically distributed around
# the diagonal, and hopefully close enough to it for the model to be useful.
```

python_scripts/metrics_regression.py
```diff
@@ -157,39 +158,56 @@
...
-# some bias.
-#
-# On this plot, we see that for the large True price values, our model tends to
-# under-estimate the price of the house. Typically, this issue arises when the
-# target to predict does not follow a normal distribution. In this case the
-# model would benefit from target transformation.
+# On these plots, we see that our model tends to under-estimate the price of the
+# house both for the lowest and large True price values. This means that the
+# residuals still hold some structure typically visible as the "banana" or
+# "smile" shape of the residual plot. This is often a clue that our model could
```
ogrisel, line 206:
```suggestion
# residuals still hold some **structure typically visible as the "banana" or
# "smile" shape of the residual plot**. This is often a clue that our model could
```

ogrisel (approved):
Some more suggestions. Thanks for the PR, this is a great methodological improvement in my opinion.

ogrisel (approved):
Thanks @ArturoAmorQ!

PR #676, opened 2022-12-06
Test scikit-learn 1.2 release candidate

lesteve:
Just to check nothing is badly broken. This is not meant to be merged.

Not sure how it will work with jupyter-cache. We may need to explicitly remove the jupyter-cache cache ...


lesteve:
OK it seems like it has run fine even when disabling the cache, so I guess we are all good here, closing

PR #678, opened 2022-12-12
Fix a small new line mistake

alerezaaa:
Hi there

First of all thanks for this course, its great

During reading the notebooks, I faced this line

and I thought maybe this highlighted line should be a new line (Due to markdown rule)

Update: I've noticed that in ReviewNB and GitHub, the newline showed correctly but in JupyterNotebook it is combined with previous line

Thanks

### DIFF
--- a/notebooks/03_categorical_pipeline.ipynb
+++ b/notebooks/03_categorical_pipeline.ipynb
@@ -425,6 +425,7 @@
     "- the original categories (before encoding) have an ordering;\n",
     "- the encoded categories follow the same ordering than the original\n",
     "  categories.\n",
+    "\n",
     "The **next exercise** highlights the issue of misusing `OrdinalEncoder` with\n",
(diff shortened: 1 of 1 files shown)


lesteve:
Thanks a lot for taking the time to open a PR!

I am not too sure whether this sentence was meant to be on the same line or not. In any case, the sentence could be made clearer.

I am tagging this for the MOOC-v4 milestone so that we don't forget to tackle it once this session (session 3) ends and before we think about starting session 4!

lesteve:
Indeed I double-checked this and this and you were right, it seems like this is a oversight from our part, I fixed it in https://github.com/INRIA/scikit-learn-mooc/commit/54ac5b119a8eefa2b0efbe6434364ebec54df63b

Let's close this PR, I decided not to modify the notebook while the MOOC was running.

Issue #681, opened 2022-12-27
[FEATURE REQUEST] Add CC-BY license to footer


darigovresearch:
Thought it may be helpful to have the fact that the course is released under a CC-BY in the footer so it's visible on every page

For example:

Brought to you under a [CC-BY license](https://github.com/INRIA/scikit-learn-mooc/blob/main/LICENSE) by [Inria Learning Lab](https://learninglab.inria.fr/), [scikit-learn @ La Fondation Inria](https://scikit-learn.fondation-inria.fr/), [Inria Academy](https://www.inria-academy.fr/formation/scikit-learn-la-boite-a-outils-de-lapprentissage-automatique/), with many thanks to the [scikit-learn](https://scikit-learn.org/) community as a whole!

lesteve:
See my comment in the associated PR: https://github.com/INRIA/scikit-learn-mooc/pull/682#issuecomment-1371871802

lesteve:
We can close this one it was merged in #682

darigovresearch:
@lesteve thanks for the help with this!

PR #682, opened 2022-12-27
feat: Adds license to jupyter-book config

darigovresearch:
Relevant to issue #681

### DIFF
--- a/jupyter-book/_config.yml
+++ b/jupyter-book/_config.yml
@@ -47,7 +47,7 @@ html:
       <div class="mooc_add">
        <a href="https://www.fun-mooc.fr/en/courses/machine-learning-python-scikit-learn">Join the full MOOC for better learning!</a>
       </div>
-      Brought to you by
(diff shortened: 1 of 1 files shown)


lesteve:
Thanks for your suggestion, I checked what the [Turing Way book](https://the-turing-way.netlify.app) is doing and it does not mention its license in its footer but only in one of its page, as we already do for this repo.

I would be in favour of not adding the licence to the footer in the but I'll let others pitch in if they have a strong opinion.

ArturoAmorQ:
I am also fine with just mentioning the license on the index, but I am open to adding it if we have a good reason why users would benefit of the increased visibility.

darigovresearch:
@lesteve & @ArturoAmorQ thanks for adding your comments. It is very common to make the license explicit (and more discoverable) in the footer as it encourages people to build on top of your work (such as translations, condensed code etc), submit corrections and code fixes.

ogrisel (approved):
I think it's a good idea to mention the license in the footer.

ArturoAmorQ (approved):
Merging, thanks!

darigovresearch:
@ArturoAmorQ thanks for the merge !

PR #689, opened 2023-02-16
MAINT Swap ax and ax.twinx in nested-cross-val figure

ArturoAmorQ:
It is more intuitive to have the outer cross-val loop to the left and the inner to the right.

In the future we can try to use a clearer notation to avoid the (miss-)interpretation that we are evaluating the model on each iteration of the inner loop.


ogrisel (approved):
Thanks for the improvement. I agree the figure is more intuitive this way.

PR #694, opened 2023-06-05
MAINT: synchronize exercises

ogrisel:
I don't think there are important changes but just in case.


ogrisel (commented):
LGTM, will merge.

PR #695, opened 2023-06-05
MAINT: update build_tools/convert-python-script-to-notebook.py to myst-parser 1.0.0

ogrisel:
I think I found the way to update the script to work with the latest version of myst-parser to avoid the `ModuleNotFoundError`.


ogrisel:
Please do not merge, I pushed the result of:

```
fix end of files.........................................................Failed
```

from the `pre-commit`. Let me exclude `ipynb` files from there and re-run.

ogrisel (commented):
LGTM now.

PR #696, opened 2023-06-05
Introduce non-linear feature engineering in more than 1D

### LINKED ISSUE #358: Sequence on non-linear feature engineering for linear models is inconsistent
- Starts with a motivating exercise in 1D => ok
- Notebook about non-linear features expansion on 1D data => ok
- Exercise on the stability of the coefficient of a linear model without non-linear pipeline???
- Quiz does not mention higher D non-linear feature preprocessing.

We could probably introduce polynomial feature interaction in a new notebook instead of the 2nd exercise.

ArturoAmorQ:
Fixes #358.

The [Exercise M4.03](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_models_sol_03.html) is part of the non-linear features section but currently it does not mention anything related to such subject. As proposed by @ogrisel, this PR introduces non-linear feature engineering in more than 1D.

I had to get rid of the boxplot mentioning the stability of the coefficients because they are not on the same scale. I can add a standard scaler step, but may make the exercise too complex.

### DIFF
--- a/python_scripts/linear_models_sol_03.py
+++ b/python_scripts/linear_models_sol_03.py
@@ -115,8 +115,8 @@ cv_results = cross_validate(
 )
 
 # %% [markdown]
-# Compute the mean and std of the MAE in grams (g) and compare with the results without
-# feature engineering.
+# Compute the mean and std of the MAE in grams (g) and compare with the results
+# without feature engineering.
 
 # %%
 # solution
--- a/python_scripts/linear_models_sol_03.py
+++ b/python_scripts/linear_models_sol_03.py
@@ -128,7 +128,7 @@ print(
 )
 
 # %% [markdown] tags=["solution"]
-# We can observe that the mean absolute error is lower and less spread with the
+# We observe that the mean absolute error is lower and less spread with the
 # enriched features. In this case the "interactions" are indeed predictive. In
 # the following notebook we will see what happens when the enriched features are
 # non-predictive and how to deal with this case.
--- a/python_scripts/linear_models_sol_03.py
+++ b/python_scripts/linear_models_sol_03.py
@@ -8,17 +8,23 @@
 # %% [markdown]
 # # 📃 Solution for Exercise M4.03
 #
-# In all previous notebooks, we only used a single feature in `data`. But we
-# have already shown that we could add new features to make the model more
-# expressive by deriving new features, based on the original feature.
+# In the previous notebook, we showed that we can add new features based on the
+# original feature to make the model more expressive, for instance x**2 or x**3.
+# In that case we only used a single feature in `data`.
 #
 # The aim of this notebook is to train a linear regression algorithm on a
-# dataset with more than a single feature.
+# dataset with more than a single feature. In such "multi-dimensional" feature
+# space we can derive new features of the form x1*x2, x2*x3, etc. Products of
+# features are usually called "interactions".
 #
-# We will load a dataset about house prices in California. The dataset consists
-# of 8 features regarding the demography and geography of districts in
-# California and the aim is to predict the median house price of each district.
-# We will use all 8 features to predict the target, the median house price.
+# Feature engineering can be an important step of a model pipeline as long as
+# the new features are expected to be predictive. For instance, think of a
+# classification model to decide if a patient has risk of developing a heart
+# disease. This would depend on the patient's Body Mass Index which is defined
+# as weight / height**2.
+#
+# We load the dataset penguins dataset. We first use a set of 3 numerical
+# features to predict the target, i.e. the body mass of the penguin.
 
 # %% [markdown]
 # ```{note}
@@ -27,10 +33,18 @@
 # ```
 
 # %%
-from sklearn.datasets import fetch_california_housing
+import pandas as pd
 
-data, target = fetch_california_housing(as_frame=True, return_X_y=True)
-target *= 100  # rescale the target in k$
+penguins = pd.read_csv("../datasets/penguins.csv")
+
+columns = ["Flipper Length (mm)", "Culmen Length (mm)", "Culmen Depth (mm)"]
+target_name = "Body Mass (g)"
+
+# Remove lines with missing values for the columns of interest
+penguins_non_missing = penguins[columns + [target_name]].dropna()
+
+data = penguins_non_missing[columns]
+target = penguins_non_missing[target_name]
 data.head()
 
 # %% [markdown]
@@ -45,7 +59,7 @@ linear_regression = LinearRegression()
 
 # %% [markdown]
 # Execute a cross-validation with 10 folds and use the mean absolute error (MAE)
-# as metric. Be sure to *return* the fitted *estimators*.
+# as metric.
 
 # %%
 # solution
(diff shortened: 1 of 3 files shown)


python_scripts/linear_models_sol_03.py
```diff
@@ -8,17 +8,23 @@
 # %% [markdown]
 # # 📃 Solution for Exercise M4.03
 #
-# In all previous notebooks, we only used a single feature in `data`. But we
-# have already shown that we could add new features to make the model more
-# expressive by deriving new features, based on the original feature.
+# In the previous notebook, we showed that we can add new features based on the
+# original feature to make the model more expressive, for instance x**2 or x**3.
```
ogrisel, line 12: The power expressions are misrendered:

image


```suggestion
# original feature to make the model more expressive, for instance `x ** 2` or `x ** 3`.
```

python_scripts/linear_models_sol_03.py
```diff
@@ -8,17 +8,23 @@
...
+# In that case we only used a single feature in `data`.
 #
 # The aim of this notebook is to train a linear regression algorithm on a
-# dataset with more than a single feature.
+# dataset with more than a single feature. In such "multi-dimensional" feature
+# space we can derive new features of the form x1*x2, x2*x3, etc. Products of
+# features are usually called "interactions".
```
ogrisel, line 18: Similar problem:

image


```suggestion
# space we can derive new features of the form `x1 * x2`, `x2 * x3`,
# etc. Products of features are usually called "non-linear or
# multiplicative interactions" between features.
```

python_scripts/linear_models_sol_03.py
```diff
@@ -8,17 +8,23 @@
...
 #
-# We will load a dataset about house prices in California. The dataset consists
-# of 8 features regarding the demography and geography of districts in
-# California and the aim is to predict the median house price of each district.
-# We will use all 8 features to predict the target, the median house price.
+# Feature engineering can be an important step of a model pipeline as long as
+# the new features are expected to be predictive. For instance, think of a
+# classification model to decide if a patient has risk of developing a heart
+# disease. This would depend on the patient's Body Mass Index which is defined
+# as weight / height**2.
```
ogrisel, line 24:
```suggestion
# as `weight / height ** 2`.
```

python_scripts/linear_models_sol_03.py
```diff
@@ -55,42 +69,65 @@
...
+)
 
-color = {"whiskers": "black", "medians": "black", "caps": "black"}
-weights.plot.box(color=color, vert=False)
-_ = plt.title("Value of linear regression coefficients")
+# %% [markdown] tags=["solution"]
+# We can observe that the mean absolute error is lower and less spread with the
+# enriched features. In this case the "interactions" are indeed predictive. In
+# the following notebook we will see what happens when the enriched features are
+# non-predictive and how to deal with this case.
```
ogrisel, line 133: We could add another question here:

Write code to answer the following question: how many features are generated at the output of the `PolynomialFeatures` feature transformer in the previous pipeline?

In addition, use set_output("pandas") to transform the first 5 rows of the dataset and look at the column names. Check that the values for the new interactions features are correct for a few of them.

python_scripts/linear_regression_non_linear_link.py
```diff
@@ -247,9 +247,9 @@
...
 # The estimator can also be configured to use a non-linear kernel. Then, it can
-# learn a prediction function that computes non-linear interaction between
-# samples for which we want to make a prediction and selected samples from the
-# training set.
+# learn a prediction function that computes non-linear relations between samples
+# for which we want to make a prediction and selected samples from the training
+# set.
 #
 # The result is another kind of non-linear regression model with a similar
 # expressivity as our previous polynomial regression pipeline:
```
ogrisel, line 255: At the very end of this notebook we could add a concluding paragraph such as:

> In this notebook we explored several ways to expand a single numerical features
> into several non-linearly derived new features to as to make our machine learning
> pipeline more expressive and less likely to underfit the data even if the last
> stage of the pipeline is a simple linear regression model.
>
> For the sake of simplicity, we introduced those transformers on a toy regression
> problem with a single input feature. However, Non-linear feature transformers such
> as `Nystroem` and others can further improve the expressiveness of machine learning
> pipelines by expanding feature sets to model non-linear feature interactions between
> features. We will explore this possibility in the next exercise.

python_scripts/linear_models_sol_03.py
ogrisel: Then we could further expand this exercise by asking:

> Now let's try to see if we can build an alternatively pipeline with
> fewer intermediate features while keeping a similar predictive power.
> To this end let's try to use the `Nystroem` transformer instead of
> `PolynomialFeatures`. Set the `kernel` parameter to `"poly"` and
> `degree` to `2`. Adjust the number of `components` to be as small as
> possible while keeping a good cross-validation performance.

ogrisel: Then in the solution we could say that `Nystroem` is a nice alternative to `PolynomialFeatures` that makes it possible to keep the memory usage of the transformed dataset under control. However, interpreting the meaning of the intermediate features can be challenging.

python_scripts/linear_models_sol_03.py
```diff
@@ -8,17 +8,23 @@
...
-# In all previous notebooks, we only used a single feature in `data`. But we
-# have already shown that we could add new features to make the model more
-# expressive by deriving new features, based on the original feature.
+# In the previous notebook, we showed that we can add new features based on the
+# original feature to make the model more expressive, for instance x**2 or x**3.
+# In that case we only used a single feature in `data`.
 #
 # The aim of this notebook is to train a linear regression algorithm on a
-# dataset with more than a single feature.
+# dataset with more than a single feature. In such "multi-dimensional" feature
```
ogrisel, line 16:
```suggestion
# dataset with more than a single feature. In such a "multi-dimensional" feature
```

python_scripts/linear_models_sol_03.py
```diff
@@ -55,42 +69,65 @@
...
-# function
-# [`df.plot.box()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.box.html)
-# to create a box plot.
-
+# Now create a pipeline using `make_pipeline` consisting of a
+# `PolynomialFeatures` and a linear regression. Set `degree=2` and
+# `interaction_only=True` to the feature engineering step. Remember not to
+# include the bias to avoid redundancies with the linear's regression intercept.
+#
+# Use the same strategy as before to cross-validate such pipeline.
```
ogrisel, line 94:
```suggestion
# Use the same strategy as before to cross-validate such a pipeline.
```

python_scripts/linear_models_sol_03.py
```diff
@@ -55,42 +70,65 @@
...
+    "Mean absolute error on testing set with interactions: "
+    f"{-cv_results['test_score'].mean():.3f} ± "
+    f"{cv_results['test_score'].std():.3f} g"
+)
 
-color = {"whiskers": "black", "medians": "black", "caps": "black"}
-weights.plot.box(color=color, vert=False)
-_ = plt.title("Value of linear regression coefficients")
+# %% [markdown] tags=["solution"]
+# We can observe that the mean absolute error is lower and less spread with the
```
glemaitre, line 131:
```suggestion
# We observe that the mean absolute error is lower and less spread with the
```

glemaitre (commented):
LGTM once @ogrisel comment are addressed.

ogrisel (approved):
LGTM, we can merge as is and open a dedicated issue to expand the exercise.

PR #698, opened 2023-06-12
MAINT Rework "Regularization of linear regression model" notebook

ArturoAmorQ:
Due to live discussions we realized that the California housing dataset contained large outliers that made the regularization unstable across cv folds. We decided to use Ames housing instead.

This PR also takes the opportunity to rework the general progression, wording and use scientific notation prints.


ogrisel (approved):
I started to do a review while experimenting locally with the notebook.

I pushed a bunch of changes directly to the PR. Let me know what you think. In particular I used symlog scale for the coefficient plots and made the ridge solver explicit because some parts of the notebook are very sensitive to it and I did not want to hide this sensitivity.

Here are the other suggestions I had in mind:

python_scripts/linear_models_regularization.py
```diff
@@ -245,48 +291,56 @@
...
 # %% [markdown]
-# We observe that scaling data has a positive impact on the test score and that
-# the test score is closer to the train score. It means that our model is less
+# We observe that scaling data has a positive impact on the test error: it is
+# now both lower and closer to the train error. It means that our model is less
 # overfitted and that we are getting closer to the best generalization sweet
 # spot.
 #
+# We also observe that fitting this pipeline no longer generates any warning
+# for any choice of the solver paramerter. Futhermore, changing the solver should
```
ogrisel, line 312: Some typos:

```suggestion
# for any choice of the solver parameter. Furthermore, changing the solver should
```

python_scripts/linear_models_regularization.py
```diff
@@ -444,8 +502,8 @@
 )
 
 # %% [markdown]
-# This range can be reduced by decreasing the spacing between the grid of
-# `alphas`.
+# This range can be reduced depending on the feature engineering and
+# preprocessing.
 #
 # In this notebook, you learned about the concept of regularization and the
 # importance of preprocessing and parameter tuning.
```
ogrisel, line 509:
```suggestion
# Here is a summary of important points highlighted in this notebook:
# - scaling features makes the effect of regularization more even: all variables
#   are regularized by comparable magnitude, which would not necessarily be the
#   case with the natural feature scales;
# - scaling features makes the numerical solvers more stable which is also helpful
#   to tune the regularization parameter more independently of the choice of the
#   solver used to fit the linear model to the training set;
# - tuning the regularization parameter of the `Ridge` estimator can be done very
#   efficiently by using the `RidgeCV` class. Wrapping it into a `cross_validate`
#   call makes it possible to assess the true generalization power of the whole
#   pipeline by including the tuning of the regularization parameter as part of the
#   learning process: this is an example of "nested cross-validation";
# - doing so makes it possible to check that the optimal value of the regularization
#   strength `alpha` is robust to a resampling of the dataset. If it wasn't the case
#   it would hint at a problem with the dataset (e.g. presence of outliers in the
#   features or the target that influence the learning process disproportionately) or
#   a bad choice of other elements of the feature engineering pipeline.
```

ogrisel (approved):
One final pass :)

PR #699, opened 2023-06-12
Add axis labels to plot about scaling

ArturoAmorQ:
Quiz M1.2 Q4 has had scores around 55% in sessions 2 and 3. We decided that maybe people thought the plot represented feature vs. target instead of feature A vs. feature B.

This PR is meant to settle the issue.


ogrisel (approved):
LGTM.

I think we could have gone with the sns scatter plot with the default way of setting the axis labels (instead of centering the axis in a cross) to avoid the unusual axis label positioning which might be a bit confusing to some students.

If you use the default seaborn scatter plot, just make sure to keep the custom (-6, 6) xlim and ylim, and instead display the grid to make it easier to see the location of the origin (0, 0) on the 2D plane.

But the current PR is good enough to me. Feel free to merge.

ArturoAmorQ:
@ogrisel I changed the plot to seaborn. It is indeed much better like this.

ogrisel:
Indeed, I also find it much clearer. Can you please add the grid? This should be doable via `ax.grid(visible=True)` if I am not mistaken.

PR #703, opened 2023-08-28
Introduce use of plot_chance_level parameter

ArturoAmorQ:
Since 1.3 we can use the `plot_chance_level` parameter for ROC and PR curves. I propose mentioning it at the end of the [Classification notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/metrics_classification.html), so that we introduce the concept of chance level using a `DummyClassifier` before showing this feature.

### DIFF
--- a/python_scripts/metrics_classification.py
+++ b/python_scripts/metrics_classification.py
@@ -394,7 +394,7 @@ _ = disp.ax_.set_title("Receiver Operating Characteristic curve")
 # performance obtained will be above this line.
 #
 # Instead of using a dummy classifier, you can use the parameter `plot_chance_level`
-# that the ROC and PR curves have implemented in scikit-learn:
+# available in the ROC and PR displays:
 
 # %%
 fig, axs = plt.subplots(ncols=2, nrows=1, figsize=(15, 7))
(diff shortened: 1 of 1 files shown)


python_scripts/metrics_classification.py
```diff
@@ -392,3 +392,32 @@
...
+    data_test,
+    target_test,
+    pos_label="donated",
+    marker="+",
+    plot_chance_level=True,
+    chance_level_kw={"color": "tab:orange", "linestyle": "--"},
+    ax=axs[1],
+)
+
+_ = plt.suptitle("PR and ROC curves")
```
glemaitre, line 423:
```suggestion
_ = fig.suptitle("PR and ROC curves")
```

python_scripts/metrics_classification.py
```diff
@@ -392,3 +392,32 @@
 # the ROC-AUC is 0.5. Indeed, we show the generalization performance of a dummy
 # classifier (the orange dashed line) to show that even the worst generalization
 # performance obtained will be above this line.
+#
+# Instead of using a dummy classifier, you can use the parameter `plot_chance_level`
+# that the ROC and PR curves have implemented in scikit-learn:
```
glemaitre, line 397:
```suggestion
# Instead of using a dummy classifier, you can use the parameter `plot_chance_level`
# available in the ROC and PR displays:
```

glemaitre (approved):
LGTM

PR #704, opened 2023-08-29
Introduce that evaluation metrics and loss are not always the same

### LINKED ISSUE #323: Find a way to introduce that the metrics and the objective function are not always the same thing in the model evaluation module
This used to be a FIXME in the markdown as reported in #319.

ArturoAmorQ:
Fixes #323.

To keep the idea of having this as extra material, it is hidden in the solution of the exercise.

Optionally we can also mention pinball loss, but I felt that we don't want to enter into the details of what different loss functions do.

### DIFF
--- a/python_scripts/metrics_sol_02.py
+++ b/python_scripts/metrics_sol_02.py
@@ -127,5 +127,5 @@ scores
 # dataset, it is true that the average MSE is lower when `loss="squared_error`,
 # whereas the average MAE is lower when `loss="absolute_error` as expected.
 #
-# If you feel like going beyond the contents of this MOOC, you can change try
-# different combinations of loss function and evaluation metrics.
+# If you feel like going beyond the contents of this MOOC, you can try different
+# combinations of loss function and evaluation metrics.
--- a/python_scripts/metrics_sol_02.py
+++ b/python_scripts/metrics_sol_02.py
@@ -95,7 +95,7 @@ scores
 # %% [markdown] tags=["solution"]
 # In the Regression Metrics notebook, we introduced the concept of loss function,
 # which is the metric optimized when training a model. In the case of a
-# linear regression, the fitting process consists in minimizing the mean squared
+# `LinearRegression`, the fitting process consists in minimizing the mean squared
 # error (MSE). Some estimators, such as the `HistGradientBoostingRegressor`, can
 # use different loss functions, to be set using the `loss` hyperparameter.
 #
@@ -120,8 +120,7 @@ for loss_func in loss_functions:
     scores["MSE"].append(f"{mse.mean():.1f} ± {mse.std():.1f}")
     scores["MAE"].append(f"{mae.mean():.1f} ± {mae.std():.1f}")
 scores = pd.DataFrame(scores)
-scores.set_index("loss", inplace=True)
-scores
+scores.set_index("loss")
 
 # %% [markdown] tags=["solution"]
 # Even if the score distributions overlap due to the presence of outliers in the
--- a/python_scripts/metrics_sol_02.py
+++ b/python_scripts/metrics_sol_02.py
@@ -93,22 +93,23 @@ scores = pd.DataFrame(scores)
 scores
 
 # %% [markdown] tags=["solution"]
-# In the Regression Metrics notebook we mentioned the concept of loss function,
-# which is the metric to be optimized when training a model. In the case of a
+# In the Regression Metrics notebook, we introduced the concept of loss function,
+# which is the metric optimized when training a model. In the case of a
 # linear regression, the fitting process consists in minimizing the mean squared
 # error (MSE). Some estimators, such as the `HistGradientBoostingRegressor`, can
 # use different loss functions, to be set using the `loss` hyperparameter.
 #
-# Notice that the evaluation metrics and the loss function are not necessarily
+# Notice that the evaluation metrics and the loss functions are not necessarily
 # the same. Let's see an example:
 
 # %%
 # solution
+from collections import defaultdict
 from sklearn.ensemble import HistGradientBoostingRegressor
 
 scoring = ["neg_mean_squared_error", "neg_mean_absolute_error"]
 loss_functions = ["squared_error", "absolute_error"]
-scores = {"loss": [], "MSE": [], "MAE": []}
+scores = defaultdict(list)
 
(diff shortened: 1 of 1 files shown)


python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
 }
 scores = pd.DataFrame(scores)
 scores
+
+# %% [markdown] tags=["solution"]
+# In the Regression Metrics notebook we mentioned the concept of loss function,
+# which is the metric to be optimized when training a model. In the case of a
```
glemaitre, line 97:
```suggestion
# which is the metric optimized when training a model. In the case of a
```

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
 }
 scores = pd.DataFrame(scores)
 scores
+
+# %% [markdown] tags=["solution"]
+# In the Regression Metrics notebook we mentioned the concept of loss function,
```
glemaitre, line 96:
```suggestion
# In the Regression Metrics notebook, we introduced the concept of loss function,
```

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
...
+# which is the metric to be optimized when training a model. In the case of a
+# linear regression, the fitting process consists in minimizing the mean squared
```
glemaitre, line 98: I find "linear regression" to be to vague here. I would instead refer to `LinearRegression` that is the scikit-learn estimator minimizing the MSE.

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
...
+# error (MSE). Some estimators, such as the `HistGradientBoostingRegressor`, can
+# use different loss functions, to be set using the `loss` hyperparameter.
+#
+# Notice that the evaluation metrics and the loss function are not necessarily
```
glemaitre, line 102:
```suggestion
# Notice that the evaluation metrics and the loss functions are not necessarily
```

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
...
+# the same. Let's see an example:
```
glemaitre, line 103: I don't know if it would be correct mentioning that we use a loss function as a proxy of the final evaluation metric that we want to optimize?

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
...
+for loss_func in loss_functions:
+    model = HistGradientBoostingRegressor(loss=loss_func)
+    cv_results = cross_validate(model, data, target, scoring=scoring)
+    mse = -cv_results["test_neg_mean_squared_error"]
+    mae = -cv_results["test_neg_mean_absolute_error"]
+    scores["loss"].append(loss_func)
+    scores["MSE"].append(f"{mse.mean():.1f} ± {mse.std():.1f}")
+    scores["MAE"].append(f"{mae.mean():.1f} ± {mae.std():.1f}")
+scores = pd.DataFrame(scores)
+scores.set_index("loss", inplace=True)
```
glemaitre, line 122: Let's not use `inplace=True` since it could be deprecated in the future.

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,41 @@
...
+# linear regression, the fitting process consists in minimizing the mean squared
+# error (MSE). Some estimators, such as the `HistGradientBoostingRegressor`, can
+# use different loss functions, to be set using the `loss` hyperparameter.
+#
+# Notice that the evaluation metrics and the loss function are not necessarily
+# the same. Let's see an example:
+
+# %%
+# solution
+from sklearn.ensemble import HistGradientBoostingRegressor
```
glemaitre, line 107:
```suggestion
from collections import defaultdict
from sklearn.ensemble import HistGradientBoostingRegressor
```

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,43 @@
...
+scores = pd.DataFrame(scores)
+scores.set_index("loss")
+
+# %% [markdown] tags=["solution"]
+# Even if the score distributions overlap due to the presence of outliers in the
+# dataset, it is true that the average MSE is lower when `loss="squared_error`,
+# whereas the average MAE is lower when `loss="absolute_error` as expected.
+# Indeed, the choice of a loss function is made depending on the evaluation
+# metric that we want to optimize for a given use case.
+#
```
ogrisel, line 131: To address Guillaume remark above, we could probably insert a paragraph here along the lines:

> Note that it is not always possible to optimize directly some metrics.
> For instance, classification models are often evaluated using metrics
> computed on hard class predictions (e.g. accuracy or F1) rather
> than from the continuous values returned by the `predict_proba`
> or `decision_function` methods. Such hard class prediction metrics
> typically evolve in a discontinuous manner when progressively changing > the internal parameters of the model during training. Therefore they
> cannot be directly optimized by learning solvers based on gradient
> descent and the likes (the details are belong the scope of this MOOC).
> So instead classifiers are typically trained by optimizing a
> loss function computed from some continuous output of the model.
> For instance `LogisticRegression` minimizes the `log_loss` applied to
> the `predict_proba` output of the model. The `log_loss` acts
> as a "surrogate loss" for discontinuous classification metrics such as
> the accuracy: by minimizing the surrogate loss, we maximize their
> accuracy. However scikit-learn does not provide surrogate losses for
> all possible classification metrics.

ogrisel (approved):
Here are some suggestions. Other than that LGTM.

python_scripts/metrics_sol_02.py
```diff
@@ -91,3 +91,62 @@
...
+# If you feel like going beyond the contents of this MOOC, you can try different
+# combinations of loss functions and evaluation metrics.
+#
+# Notice that there are some metrics that cannot be directly optimized by
+# optimizing a loss function. This is the case for metrics that evolve in a
+# discontinuous manner with respect to the internal parameters of the model, as
+# learning solvers based on gradient descent or similar optimizers require
+# continuity (the details are belong the scope of this MOOC).
```
glemaitre, line 139:
```suggestion
# continuity (the details are beyond the scope of this MOOC).
```

PR #708, opened 2023-09-01
Synchronize quiz

ogrisel:
From the private repo with solutions.


ArturoAmorQ (approved):
Should be good as it is. Thanks @ogrisel, merging!

PR #702, opened 2023-07-13
MAINT Use class `ValidationCurveDisplay`

ArturoAmorQ:
[ValidationCurveDisplay](https://scikit-learn.org/dev/modules/generated/sklearn.model_selection.ValidationCurveDisplay.html) is available since 1.3. This PR makes use of this new class for easing the plotting a little bit.

Notice this requires updating the requirement to `scikit-learn = 1.3`.

PR #709, opened 2023-09-01
MAINT Update scikit-learn to v 1.3

ArturoAmorQ:
PR #702 introduced `ValidationCruveDisplay` which requires v >= 1.3. This PR updates the requirements and, as a side effect, formats the code on those files.


ogrisel (approved):
Thanks very much!

Issue #707, opened 2023-08-31
Non linear feature engineering for logistic regression


ogrisel:
As a follow up for #701, I suggest that:

- We replace the notebook currently named "Beyond linear separation in classification" by a new notebook named "Non-linear feature engineering for Logistic Regression"
- In this notebook we reuse the same 2D synthetic moons and Gaussian quantiles datasets
- We start with a logistic regression and shows that it underfits
- Then we build more and more complex pipelines with different preprocessors:
- `KBinsDiscretizer`
- `SplineTransformer`

- We observe that those transformers do axis-aligned non linear transformations that lead to axis aligned classification decision boundaries,
- We explore modeling multiplicative interactions between the derived features with
- `KBinsDirectizer` with sparse output followed by `PolynomialFeatures(degree=2, interaction_only=True)`
- `SplineTransformer` followed by `Nystroem` (either with kernel="rbf" and a good value of `gamma` or `kernel="poly"` and `degree=2`)

Then we add a new exercise with:

- The half moons dataset only
- `SVC(kernel="linear")` (this should give similar underfitting results as logistic regression from the previous notebook
- then ask the user to try:
    - `make_pipeline(Nystroem(kernel="rbf", gamma=some_gamma, n_components=300), SVC(kernel="linear")`
    - SVC(kernel="rbf", gamma=some_gamma)
- the results should be similar

Then we can optionally suggest to try `MLPClassifier` on this dataset to get somewhat similar results.

glemaitre:
working on this one

PR #711, opened 2023-09-01
ENH new non-linear engineering for LogisticRegression

glemaitre:
Towards #707


ogrisel:
@glemaitre @ArturoAmorQ I think this is ready for review. I plan to open dedicated PRs to address the remaining items of #707.

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+#
+# We use matplotlib to visualized the data points in both datasets:
+
+# %%
+import matplotlib.pyplot as plt
+import seaborn as sns
+import warnings
+
+
+warnings.filterwarnings("ignore", module="seaborn")
```
ArturoAmorQ, line 89: I noticed seaborn is raising FutureWarning in several places inside the notebook. I am wondering if the best solution is to ping to a previous version. But at least we should mention why we are adding this line, WDYT?

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+            (data_moons, target_moons),
+            (data_gauss, target_gauss),
+            (data_xor, target_xor),
+        ],
+    ):
+        model.fit(data, target)
+        DecisionBoundaryDisplay.from_estimator(
+            model,
+            data,
+            response_method="predict_proba",
```
ArturoAmorQ, line 150: We introduce the concept of `predict_proba` [here in M5](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_classification.html?highlight=predict_proba). In my opinion it would be better to keep the style for the decision boundary display as introduced [at the beginning of M4](https://inria.github.io/scikit-learn-mooc/python_scripts/logistic_regression.html), to rather have the discussion focused on the kernels.

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+# for linear models.
+
+# %% [markdown]
+#
+# ## Engineering non-linear features
+#
+# As we did for the linear regression models, we will attempt to build a more
+# expressive machine learning pipeline by leveraging non-linear feature
+# engineering. We will use the techniques as for linear regression models:
+# binning, splines, polynomial features, and kernel approximation.
```
ArturoAmorQ, line 210: In the linear regression module we did not use splines. Should we?

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+# %%
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from sklearn.linear_model import LogisticRegression
+
+logistic_regression = make_pipeline(StandardScaler(), LogisticRegression())
+logistic_regression
+
+# %%
+axs = plot_decision_boundary(logistic_regression)
```
ArturoAmorQ, line 194: Can we add a suptitle to distinguish each plot from each other?

Also we may need some horizontal spacing between subplots, as they are overlapping in the rendering.

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+
+# %% [markdown]
+#
+# This confirms that it is not possible to separate the two classes with a
+# linear model. On each plot we see a **significant number of misclassified
+# samples on the training set**! This is a typical example of **underfitting**
+# for linear models.
```
ArturoAmorQ, line 201: We can add a short comment on the somewhat unexpected behavior of the XOR plot, similar to "the last plot performs specially bad, as it tries to correctly classify most of the samples in the positive class, even if that means incorrectly identifying every sample of the negative class."

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+# the moons dataset. If we use too many knots, the model will overfit the data.
+#
+# However, as for the binning transformation, the model still fails to separate
+# the data for the XOR dataset, irrespective of the number of knots, for the
+# same reasons: the spline transformation is a feature-wise transformation and
+# thus cannot capture interactions between features.
+
+# %% [markdown]
+#
+# Modeling non-additive feature interactions
```
ArturoAmorQ, line 273:
```suggestion
# ## Modeling non-additive feature interactions
```

jupyter-book/_toc.yml
```diff
@@ -102,6 +102,7 @@ parts:
     - file: python_scripts/linear_regression_non_linear_link
     - file: python_scripts/linear_models_ex_02
     - file: python_scripts/linear_models_sol_02
+    - file: python_scripts/linear_models_feature_engineering_classification.py
     - file: python_scripts/logistic_regression_non_linear
```
ArturoAmorQ, line 106: This new notebook makes the logistic_regression_non_linear notebook obsolete, shall we remove it?

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+# %% [markdown]
+#
+# # Non-linear feature engineering for Logistic Regression
+#
+# Linear classification models can also be used for classification on
+# non-linearly separable datasets. To do so, we transform the original feature
+# feature space into a new feature space using a non-linear transformation. The
+# resulting feature space has typically more dimensions than the original
+# feature space, and as a result, the linear model can separate the data more
+# easily.
```
ArturoAmorQ, line 17:
```suggestion
# In the slides at the beginning of the module we mentioned that linear
# classification models are not suited to non-linearly separable data.
# Nevertheless, one can still use feature engineering as previously done for
# regression models to overcome this issue. To do so, we use non-linear
# transformations that typically map the original feature space into a higher
# dimension space, where the linear model can separate the data more easily.
```

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,437 @@
...
+            y=feature_names[1],
+            hue=target,
+            palette=["tab:red", "tab:blue"],
+            ax=ax,
+        )
+    return axs
+
+
+# %% [markdown]
+#
```
ArturoAmorQ, line 181:
```suggestion
```

ArturoAmorQ (commented):
Thanks for the PR @glemaitre! Here is a first batch of comments.

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: I insert those blank lines intentionally at the beginning of the markdown cells to make it easier to just press alt-q in VS Code to rewrap the markdown content of the cell without having to select only the lines I want.

If I don't do this, re-wrapping by pressing alt-q on this cell would yield the following invalid markdown:

```
# %% [markdown] Now let's define our logistic regression model and plot its decision
# boundary on the three datasets:
```

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: I think we can just add a quick note to explain that the background color of this plot represents the confidence level for each class (white meaning "unsure about which class to choose") using the `predict_proba` method and that more details about the use and the meaning of the `predict_proba` method will be given in the next module.

ogrisel: I think this is really important to show the impact of the regularization on the confidence of the model (in the following exercise) and therefore I think it's important to introduce this way of plotting the decision boundaries as soon as possible (before the exercise, not to introduce too many new concept at once in the exercise itself).

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: I thought we did... Maybe we we should (quickly). Right after the polynomial features and before the kernel.

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: I agree, those warnings are annoying. But rather than pinning seaborn, we might have to pin an older version of pandas. Or both.

ogrisel: Or maybe we can put this call to `warnings.filterwarnings("ignore", module="seaborn")` outside of the notebooks themselves, maybe in `conf.py` or something similar if sphinx or jupyter book allow for this kind of thing.

jupyter-book/_toc.yml
ogrisel: I wanted to convert it into an exercise (about the relation ship between nystrom + linear SVC on one hand and SVC with kernel on the other hand) but I plan to do that in a follow-up PR.

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: Also note: the extra blank space is not visible for the end-user, so we don't need to be consistent everywhere. It's just a convenience when editing with VS Code.

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: The titles have been added in 1338b95.

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: I removed the warnings filter in 1338b95.

python_scripts/linear_models_feature_engineering_classification.py
ogrisel: Done in 0a53ccc.

ogrisel:
@ArturoAmorQ I think all the comments that are directly related to this particular notebook have been addressed and this PR is ready for final review and merge.

I will open follow-up PRs to update the other notebooks (e.g. transform the kernel notebook into an exercise (https://github.com/INRIA/scikit-learn-mooc/pull/711#discussion_r1319817018) and introduce splines in the previous notebook(s) where appropriate (https://github.com/INRIA/scikit-learn-mooc/pull/711#discussion_r1319796526).

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,460 @@
...
+)
+data_moons, target_moons = moons[feature_names], moons[target_name]
+
+# %% [markdown]
+#
+# The second dataset is called the "Gaussian quantiles" dataset as the data
+# points are sampled from the same (Gaussian) distribution in a 2D space with a
+# higher density of points in the center. The points closest to the center are
+# assigned to the class 1 while the points in the outer edges are assigned to
+# the class 0.
```
ArturoAmorQ, line 47:
```suggestion
# The second dataset is called the "Gaussian quantiles" dataset as all data
# points are sampled from a 2D Gaussian distribution regardless of the class.
# The points closest to the center are assigned to the class 1 while the points
# in the outer edges are assigned to the class 0, resulting in concentric
# circles.
```

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,460 @@
...
+)
+classifier
+# %%
+plot_decision_boundary(classifier, title="Binning + Nystroem classifier")
+
+# %% [markdown]
+#
+# It is interesting to observe that this model is still piecewise constant with
+# axis-aligned decision boundaries everywhere, but it can now successfully deal
+# with the XOR problem thanks to the the second step of the pipeline that can
```
ArturoAmorQ, line 402:
```suggestion
# with the XOR problem thanks to the second step of the pipeline that can
```

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,460 @@
...
+#
+# ## Summary and take-away messages
+#
+# - Linear models such as logistic regression can be used for classification on
+#   non-linearly separable datasets by leveraging non-linear feature
+#   engineering.
+# - Transformers such as `KBinsDiscretizer` and `SplineTransformer` can be used
+#   to engineer non-linear features independently for each original feature.
+# - As a result, these transformers cannot capture interactions between the
+#   orignal features (and the would fail on the XOR classification task).
```
ArturoAmorQ, line 437:
```suggestion
#   orignal features (and then would fail on the XOR classification task).
```

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,460 @@
...
+# - They also favor axis-aligned decision boundaries, in particular in the low
+#   density regions of the feature space (axis-aligned extrapolation).
+# - Transformers such as `PolynomialFeatures` and `Nystroem` can be used to
+#   engineer non-linear features that capture interactions between the original
+#   features.
+# - It can be useful to combine several feature engineering transformers in a
+#   single pipeline to build a more expressive model, for instance to favor
+#   axis-aligned extrapolation while also capturing interactions.
+# - In particular, if the original dataset has both numerical and categorical
+#   features, it can be useful to apply inning or spline transformation to the
```
ArturoAmorQ, line 449:
```suggestion
#   features, it can be useful to apply binning or a spline transformation to the
```

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,460 @@
...
+#   numerical features and one-hot encoding to the categorical features. Then,
+#   the resulting features can be combined with a kernel approximation to model
+#   interactions between numerical and categorical features. This can be
+#   achieved with the help of the `ColumnTransformer` transformer.
```
ArturoAmorQ, line 453:
```suggestion
#   achieved with the help of `ColumnTransformer`.
```

python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -0,0 +1,460 @@
...
+        datasets,
+    ):
+        model.fit(data, target)
+        DecisionBoundaryDisplay.from_estimator(
+            model,
+            data,
+            response_method="predict_proba",
+            plot_method="pcolormesh",
+            cmap="RdBu",
+            alpha=0.8,
```
ArturoAmorQ, line 167:
```suggestion
            alpha=0.8,
            # Setting vmin and vmax to the extreme values of the probability to
            # ensure that 0.5 is mapped to white (the middle) of the blue-red
            # colormap.
```

ogrisel:
Thanks very much for the fixes and improvements Arturo.

ArturoAmorQ:
Thanks @glemaitre and @ogrisel :) I am merging this one.

PR #701, opened 2023-06-26
Rework ordering of linear models module

ArturoAmorQ:
Fixes #697. Fixes #641.

This PR addresses al the point in #697 except for the point:
> rename "Beyond linear separation in classification"

This will be addressed in a different PR as I would like to fit the narrative to the new placing of this notebook.

Furthermore, we can consider moving Q2 from M4.05
> The decision boundaries of a logistic regression:
    a) uses only one of the input features
    b) uses a combination of the input features
    c) separates a single class

out of the regularization lesson and into the Quiz M4.02, i.e. in the lesson on intuitions.

PR #713, opened 2023-09-08
MAINT fix narrative according to new order of M4

ArturoAmorQ:
Follows #701, were we removed one exercise. This PR removes the narrative references made to it.

### DIFF
--- a/python_scripts/linear_models_sol_02.py
+++ b/python_scripts/linear_models_sol_02.py
@@ -90,7 +90,9 @@ print(
 # Now create a pipeline using `make_pipeline` consisting of a
 # `PolynomialFeatures` and a linear regression. Set `degree=2` and
 # `interaction_only=True` to the feature engineering step. Remember not to
-# include the bias to avoid redundancies with the linear regression intercept.
+# include a "bias" feature (that is a constant-valued feature) to avoid
+# introducing a redundancy with the intercept of the subsequent linear
+# regression model.
 #
 # Use the same strategy as before to cross-validate such a pipeline.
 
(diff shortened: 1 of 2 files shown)


python_scripts/linear_models_sol_02.py
```diff
@@ -90,7 +90,7 @@
 # Now create a pipeline using `make_pipeline` consisting of a
 # `PolynomialFeatures` and a linear regression. Set `degree=2` and
 # `interaction_only=True` to the feature engineering step. Remember not to
-# include the bias to avoid redundancies with the linear's regression intercept.
+# include the bias to avoid redundancies with the linear regression intercept.
```
ogrisel, line 93:
```suggestion
# include a "bias" feature (that is a constant-valued feature) to avoid
# introducing a redundancy with the intercept of the subsequent linear
# regression model.
```

python_scripts/linear_regression_non_linear_link.py
```diff
@@ -212,9 +210,9 @@
 
 # %% [markdown]
 # The last possibility is to make a linear model more expressive is to use a
-# "kernel". Instead of learning a weight per feature as we previously
-# emphasized, a weight will be assigned to each sample. However, not all samples
-# will be used. This is the base of the support vector machine algorithm.
+# "kernel". Instead of learning one weight per feature as we previously did, a
+# weight is assigned to each sample. However, not all samples are used. This is
+# the base of the support vector machine algorithm.
```
ogrisel, line 215:
```suggestion
# weight is assigned to each sample. However, not all samples are used: some
# redundant data point of the training set are assigned a weight of 0 so
# that they do no influence the model's prediction function. This is the
# main intuition of the support vector machine algorithm.
```

ogrisel (approved):

PR #710, opened 2023-09-01
ENH use probability estimates in DecisionBoundaryDisplay

glemaitre:

PR #715, opened 2023-09-12
ENH Introduce predict_proba earlier in the module

ArturoAmorQ:
This is a first step to introducing #710 and #711 in a more smooth way.


ogrisel (approved):
Thanks @ArturoAmorQ. We could even add the plot with `response_method="predict_proba"` at then end of the notebook rather than just telling the reader to try.

As you wish.

PR #717, opened 2023-09-13
ENH Expand contents related to Nystroem

ArturoAmorQ:
Addresses suggestions in [this commment](https://github.com/INRIA/scikit-learn-mooc/pull/696#discussion_r1309981324) by @ogrisel.

Side effects:
- adds reminder on reverting the sign in "neg_mean_absolute_error";
- fixes reference "in the following notebook" which is no longer true after the reordering;
- changes the behavior of Nystroem on the [Linear regression for a non-linear features-target relationship notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/linear_regression_non_linear_link.html) to use a polynomial kernel, as it seems more pertinent to the contents of the notebook.

### DIFF
--- a/python_scripts/linear_models_ex_02.py
+++ b/python_scripts/linear_models_ex_02.py
@@ -125,5 +125,9 @@ data.head()
 # %%
 # Write your code here.
 
+# %% [markdown]
+# How does the mean and std of the MAE for the Nystroem pipeline with optimal
+# `n_components` compare to the other previous models?
+
 # %%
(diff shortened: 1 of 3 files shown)


ogrisel (approved):
Thanks @ArturoAmorQ. I pushed 2 commits to further expand on this, and in particular add a final analysis.

I decided to reduce the parameter range. Otherwise the oscillating validation curve is misleading (because of the size of the CV error bars).

PR #721, opened 2023-09-20
Synchronize notebooks by running 'make notebooks'

ogrisel:
Apparently, it's been a while since we hadn't done it :)


ArturoAmorQ (approved):
Should be good to go. Thanks @ogrisel :)

PR #727, opened 2023-10-05
MAINT Fix pandas version to 2.0 to avoid seaborn warning

ArturoAmorQ:
The version of pandas 2.1 is conflicting with seaborn. This is a quick fix until the issue is addressed on their side.

I did not modify check_env.py to keep the minimal required version of pandas to 1.0


ogrisel (approved):

lesteve:
FYI seaborn 0.13 has been released 29 September I think this should allow to unpin pandas.

PR #730, opened 2023-10-10
ENH Emphasize discussion on multi-class classification in tree notebook

ArturoAmorQ:
This PR reworks the general wording to avoid redundant text and prefer verbs in present mode.
It also adds a plot of `predict_proba` per class, as inspired by the [Plot classification probability example](https://scikit-learn.org/dev/auto_examples/classification/plot_classification_probability.html)


ogrisel (approved):
I pushed some commits to adjust the colormaps to:

- avoid using a binary Red Blue colormap for multiclass decision boundary,
- use a continuous viridis for the predict_proba plots because 0.5 is not meaningful any more in a one vs rest setting.

Below are some more suggestions, but beyond this, LGTM.

python_scripts/trees_sol_01.py
```diff
@@ -114,3 +125,65 @@
...
+# `predict_proba` for a multiclass problem using `DecisionBoundaryDisplay`,
+# except that for a K-class problem you have K probability outputs for each
+# data point. Visualizing all these on a single plot can quickly become tricky
+# to interpret. It is then common to instead produce K separate plots, one for
+# each class, in a one-vs-rest (or one-vs-all) fashion.
+#
+# For example, in the plot below, the first plot on the left shows in red the
+# certainty on classifying a data point as belonging to the "Adelie" class. In
+# the same plot, the blue color represents the certainty of **not** belonging to
+# the "Adelie" class. The same logic applies to the other plots in the figure.
```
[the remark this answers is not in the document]
ogrisel: Indeed this paragraph needs to be updated with the new colors (e.g. bright yellow vs dark purple). What I mean is that the chance level for a one vs rest binary classification problem that comes from a multi-class classification problem is almost never at 0.5. So using a colormap with a neutral white at 0.5 might give a false impression.

When we do one-vs-rest, we do not threshold the value of predict_proba at 0.5 but instead take argmax of the concatenation of the one-vs-rest predict_proba across the classes dimension.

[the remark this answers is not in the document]
ArturoAmorQ: Does it make sense to keep the `colorbar` at the bottom of the plot in this case?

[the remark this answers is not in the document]
ogrisel: I believe so.

python_scripts/trees_classification.py
```diff
@@ -92,13 +99,28 @@
 print(f"Accuracy of the LogisticRegression: {test_score:.2f}")
 
 # %% [markdown]
-# Unlike linear models, decision trees are non-parametric models: they are not
-# controlled by a mathematical decision function and do not have weights or
+# Unlike linear models, the decision rules for the decision trees are not
+# controlled by a mathematical decision function and do not have weights or an
 # intercept to be optimized.
```
ogrisel, line 104: This sentence is still not correct: a decision tree prediction function can still be written an a mathematical expression (e.g. a sum of products with boolean indicator terms for the split nodes and weights for the leaf nodes). However the number of terms and trainable "weights" is not fixed ahead of time as for linear models.

The following is more correct:


```suggestion
# Unlike linear models, the decision rule for the decision tree is not
# controlled by a simple linear combination of weights and feature values.
```

python_scripts/trees_classification.py
```diff
@@ -92,13 +99,28 @@
...
 #
-# Indeed, decision trees will partition the space by considering a single
-# feature at a time. Let's illustrate this behaviour by having a decision tree
-# make a single split to partition the feature space.
+# Instead, the decision rules of trees can be defined in terms of
+# - the feature index,
+# - the threshold values used at each split,
+# - the respective value to predict.
```
ogrisel, line 109:
```suggestion
# - the feature index used at each split node of the tree,
# - the threshold value used at each split node,
# - the value to predict at each leaf node.
```

python_scripts/trees_classification.py
```diff
@@ -92,13 +99,28 @@
...
+#
+# As the number of splitting nodes may grow when changing the number of data
+# points in the training set, the number of trainable decision rules grows as
+# well. This is why we say that decision trees are non-parametric models.
```
ogrisel, line 113:
```suggestion
# As the number of split and leaf nodes may grow when changing the number of data
# points in the training set, the number of adjustable components taking part in
# the decision rule would grow as well. This is why we say that decision trees are
# non-parametric models: the number of trainable parameters depends on the 
# size and regularity of the training set.
```

python_scripts/trees_classification.py
```diff
@@ -92,13 +99,28 @@
...
+#
+# Decision trees partition the feature space by considering a single feature at
+# a time. The number of splits depends on both the hyperparameters and the
+# number of data points in the training set: the larger the training set, the
+# more splits can be considered by the model (assuming a flexible enough choice
+# for the hyperparameters).
```
ogrisel, line 119:
```suggestion
# number of data points in the training set: the more flexible the
# hyperparameters and the larger the training set, the more splits can be
# considered by the model.
```

python_scripts/trees_sol_01.py
```diff
@@ -114,3 +132,68 @@
...
+    if k == 0:
+        axs[k].set_ylabel("Culmen Depth (mm)")
+    idx = target_test == tree.classes_[k]
+    axs[k].scatter(
+        data_test["Culmen Length (mm)"].loc[idx],
+        data_test["Culmen Depth (mm)"].loc[idx],
+        marker="o",
+        c="w",
+        edgecolor="k",
+    )
```
ogrisel, line 186: Maybe this makes it easier to relate to what we say in the last paragraph.

```suggestion
for class_of_interest in range(n_classes):
    axs[class_of_interest].set_title(f"Class {tree.classes_[class_of_interest]}")
    imshow_handle = axs[class_of_interest].imshow(
        probas[:, class_of_interest].reshape((100, 100)),
        extent=(30, 60, 10, 23),
        vmin=0.0,
        vmax=1.0,
        origin="lower",
        cmap="viridis",
    )
    axs[class_of_interest].set_xlabel("Culmen Length (mm)")
    if class_of_interest == 0:
        axs[class_of_interest].set_ylabel("Culmen Depth (mm)")
    idx = target_test == tree.classes_[class_of_interest]
    axs[class_of_interest].scatter(
        data_test["Culmen Length (mm)"].loc[idx],
        data_test["Culmen Depth (mm)"].loc[idx],
        marker="o",
        c="w",
        edgecolor="k",
    )
```

ogrisel (approved):
A final batch of feedback. Otherwise LGTM. The rendering looks good.

PR #731, opened 2023-10-17
ENH Convert some of the Wrap-up M4 content into exercise

### LINKED ISSUE #707: Non linear feature engineering for logistic regression

ArturoAmorQ:
Fixes #707.

Follows #711, which made the `logistic_regression_non_linear` notebook redundant. This PR creates an exercise to show the use of feature engineering using a more realistic dataset. In particular, demonstrates feature interaction when using one-hot encoding.

**Note**: I had to build the exercises from the solutions to correctly render the index.


python_scripts/linear_models_ex_03.py
```diff
@@ -14,69 +14,116 @@
...
-from sklearn.preprocessing import StandardScaler
-from sklearn.linear_model import LogisticRegression
+# preprocessor.fit(data)
+# feature_names = (
+#     preprocessor.named_transformers_["onehotencoder"].get_feature_names_out(
+#         categorical_columns
+#     )
+# ).tolist()
+# feature_names += numerical_columns
+# feature_names
```
ArturoAmorQ, line 90: For info: I had to comment this lines by hand as it was rising a flake8 error `F821 undefined name` when building the exercise from the solution.

We may need to think of a better way to avoid this situation in the future.

ogrisel:
Similarly to the private review made on the wrap up quiz, I am worried that fitting such a pipeline might require too much RAM and CPU on jupyterhub and as a result might crash or be painfully slow to execute. We should either trim the number of categorical features using `min_frequency` or `max_categories` in `OneHotEncoder` or alternatively use a `Nystroem` approximation with `n_components` between 100 and 1000.

Furthermore, I would also replace standard scaling by `SplineTransformer` to get a more expressive model.

python_scripts/linear_models_ex_03.py
ArturoAmorQ: The issue here is that `preprocessor` is defined in the solution but not in the exercise. So I think it will raise an error when building the jupyterbook.

ogrisel: Indeed. We can keep it that way then.

Maybe add a comment to state to reuse the `preprocessor` variable defined in the solution of the previous question.

ArturoAmorQ:
> Furthermore, I would also replace standard scaling by `SplineTransformer` to get a more expressive model.

I decided not to do so because feature engineering actually degraded the performance of the spline model.

python_scripts/linear_models_ex_03.py
```diff
@@ -5,7 +5,7 @@
 #       extension: .py
 #       format_name: percent
 #       format_version: '1.3'
-#       jupytext_version: 1.14.5
+#       jupytext_version: 1.15.2
```
ArturoAmorQ, line 8:
```suggestion
#       jupytext_version: 1.14.5
```

ogrisel:
Hum, it seems that I broken the jupyter preview...

ArturoAmorQ:
I will have to merge this PR in its current state to possibly debug the synchro of notebooks and their impact on FUN. We can always iterate on it on future PRs.

PR #734, opened 2023-10-20
MAINT use actions/cache@v3 instead of v2

ogrisel:
When running the gh-pages deployment one gets the following warning:

```
deploy-gh-pages
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/cache@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/

```


ogrisel:
The jupyter preview does not point to the right PR. I am working on a fix.

ogrisel:
It seems to work as expected. Let's merge and see if it fixes the problem when having current workflow running in different PRs.

PR #739, opened 2023-10-25
ENH Mention scaling behavior of binning and splines

ArturoAmorQ:
Follows #711.

In the M4 wrap-up quiz we ask students to use splines instead of standard scaling.
This PR sets a bit of context on why that can be done.

Side effect: A bit of wording of the previous content.

### DIFF
--- a/python_scripts/linear_models_feature_engineering_classification.py
+++ b/python_scripts/linear_models_feature_engineering_classification.py
@@ -279,15 +279,21 @@ plot_decision_boundary(classifier, title="Spline classifier")
 # We can see that the decision boundary is now smooth, and while it favors
 # axis-aligned decision rules when extrapolating in low density regions, it can
 # adopt a more curvy decision boundary in the high density regions.
-#
-# Note however, that the number of knots is a hyperparameter that needs to be
-# tuned. If we use too few knots, the model would underfit the data, as shown on
-# the moons dataset. If we use too many knots, the model would overfit the data.
-#
 # However, as for the binning transformation, the model still fails to separate
 # the data for the XOR dataset, irrespective of the number of knots, for the
 # same reasons: **the spline transformation is a feature-wise transformation**
 # and thus **cannot capture interactions** between features.
+#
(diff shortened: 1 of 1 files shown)


python_scripts/linear_models_feature_engineering_classification.py
```diff
@@ -279,15 +279,21 @@ def plot_decision_boundary(model, title=None):
...
+#
+# Take into account that the number of knots is a hyperparameter that needs to be
+# tuned. If we use too few knots, the model would underfit the data, as shown on
+# the moons dataset. If we use too many knots, the model would overfit the data.
+#
+# ```{note}
+# Notice that `KBinsDiscretizer` and `SplineTransformer` do not require
+# additional scaling. Indeed, they can replace the scaling step for numerical
+# features, as the former maps the features into the range [0, `n_bins`] and the
+# later maps them into the range [0, 1].
```
ogrisel, line 295:
```suggestion
# Notice that `KBinsDiscretizer(encode="onehot")` and `SplineTransformer` do not
# require additional scaling. Indeed, they can replace the scaling step for numerical
# features: they both create features with values in the [0, 1] range.
```

Note that `KBinsDiscretizer` does the one hot encoding by default:

```python
In [1]: from sklearn.preprocessing import KBinsDiscretizer

In [2]: KBinsDiscretizer(n_bins=5)
Out[2]: KBinsDiscretizer()

In [3]: KBinsDiscretizer(n_bins=5).get_params()
Out[3]: 
{'dtype': None,
 'encode': 'onehot',
 'n_bins': 5,
 'random_state': None,
 'strategy': 'quantile',
 'subsample': 'warn'}
```

However we could make that more explicit by using:


```python
classifier = make_pipeline(
    KBinsDiscretizer(n_bins=5, encode="onehot"), LogisticRegression()
)
```

at line 238 of this notebook instead of using the implicit default.

ogrisel (approved):
One the following comment is addressed, LGTM!

ogrisel (approved):

PR #741, opened 2023-10-26
FIX jupyter book preview in PRs

ogrisel:
I might have broken the jupyter book preview in PRs when merging #734.


ogrisel:
Let's merge and open another debug PR to check it works.

PR #743, opened 2023-10-26
FIX Notebooks not updated by `make notebooks`

ArturoAmorQ:
As reported by @lfarhi in RL, some notebooks were not updated by `make notebooks` in #740 and I had to create them by hand running `jupytext --to notebook`:

- linear_models_feature_engineering_classification.ipynb was introduced in #711 and therefore not previously tracked
- linear_regression_non_linear_link.ipynb
- logistic_regression.ipynb

This is probably a bug in the makefile. The list may not be extensive and requires more investigation.


ArturoAmorQ:
@ogrisel can you try to run make on your machine ? Maybe it is my config that is bugging.

PR #745, opened 2023-10-29
DEBUG jupyter book preview in PRs

ogrisel:
Let's check if #741 works.


ogrisel:
This still does not work.

PR #753, opened 2023-11-19
FIX fix description to be aligned with figure legend

glemaitre:
Fixing the typo reported in:
https://mooc-forums.inria.fr/moocsl/t/suspected-typo/15589


ArturoAmorQ (approved):
Thanks @glemaitre!

I think it should be safe to fix the typo for new users on FUN, even if we don't ask old users to reload the notebook.

PR #757, opened 2023-11-29
MAINT Update full index

lesteve:
Realised this during the Inria Academy session.

- fixed the `Makefile` that had the old python script name
- temporarily pin myst_parser since myst_parser 0.18 (released June 2022 ...) removed a function that is used in build_tools/generate-index.py. Finding an alternative way to do the same thing is left for a future PR
- update full-index.ipynb

This does not affect the MOOC and is used only during live teaching so this can be merged without sweating too much about it.


lesteve:
I double-checked and this seems consistent with the jupyter-book, merging

PR #760, opened 2024-02-06
ENH Improve wording in stratification notebook

ArturoAmorQ:
During a formation at Inria Academy we noticed that this notebook never really justifies why stratification is important. This PR adds a couple of paragraphs to better motivate the reason why a simple `KFold` with shuffling is not a good enough practice.

It also takes the opportunity to implement verbs in present mode and improve general wording.

NB. I think this PR is safe to merge at it does not change the overall experience of the mooc.

### DIFF
--- a/python_scripts/cross_validation_stratification.py
+++ b/python_scripts/cross_validation_stratification.py
@@ -263,8 +263,6 @@ _ = plt.title("Test set")
 # performance metrics depend on the proportion of the positive class, as we will
 # see in a future notebook.
 #
-# In conclusion, it is a good practice to use stratification within the
-# cross-validation framework when dealing with a classification problem,
-# especially for datasets with imbalanced classes or when the class distribution
-# is crucial for model evaluation. This approach provides a more realistic
-# assessment of the model's ability to generalize across the entire dataset.
+# The interested reader can learn about other stratified cross-validation
+# techniques in the [scikit-learn user
+# guide](https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation-iterators-with-stratification-based-on-class-labels).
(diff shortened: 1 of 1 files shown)


python_scripts/cross_validation_stratification.py
```diff
@@ -255,5 +256,14 @@
...
-# cross-validation framework when dealing with a classification problem.
+# In other words, `StratifiedKFold` maintains the original distribution of
+# classes in each fold, ensuring that each fold is a good representative of the
+# whole dataset. This can have a particular impact when using performance
+# metrics that depend on the proportion of the positive class, as we
+# will see in a future notebook.
+#
+# In conclusion, it is a good practice to use stratification within the
+# cross-validation framework when dealing with a classification problem,
+# especially for datasets with imbalanced classes or when the class distribution
```
glemaitre, line 267: I am not really sure of the conclusion. To me, I'm thinking more about the following two aspects:

- if target labels are ordered or grouped, then stratification allows to overcome this issue if you forget to shuffle as in the k-fold case;
- if the sample size is limited or small (and data are shuffled), then taking a stratified fold ensure similar train/test distribution compare to the uniform sampling.

And thus overcoming the both above issues make that the evaluation is closer to reality.

glemaitre (commented):
Otherwise LGTM.

python_scripts/cross_validation_stratification.py
ArturoAmorQ: I am not entirely convinced on the rewording I made, but I think I addressed your points in [42c9775](https://github.com/INRIA/scikit-learn-mooc/pull/760/commits/42c9775ccc77709e2679efdbe7be8b718365cde1)

glemaitre (approved):
LGTM.

PR #763, opened 2024-02-21
ENH Rework narrative of GBDT notebook

ArturoAmorQ:
Originally I wanted to rework only the wording, as well explaining the GBDT algo is the cornerstone for understanding HGBT. But then I decided to factorize the code by introducing a helper function to keep the focus on the narrative other than the code.

### DIFF
--- a/python_scripts/ensemble_gradient_boosting.py
+++ b/python_scripts/ensemble_gradient_boosting.py
@@ -87,7 +87,7 @@ target_test_predicted = tree.predict(data_test)
 # %%
 def plot_decision_tree_with_residuals(y_train, y_train_pred, y_test_pred):
     """Plot the synthetic data, predictions, and residuals for a decision tree.
-    Handles are used to create custom legends for the plot."""
+    Handles are returned to allow custom legends for the plot."""
     _fig_, ax = plt.subplots()
     # plot the data
     sns.scatterplot(
--- a/python_scripts/ensemble_gradient_boosting.py
+++ b/python_scripts/ensemble_gradient_boosting.py
@@ -86,7 +86,8 @@ target_test_predicted = tree.predict(data_test)
 
 # %%
 def plot_decision_tree_with_residuals(y_train, y_train_pred, y_test_pred):
-    # Create a plot and get the Axes object
+    """Plot the synthetic data, predictions, and residuals for a decision tree.
+    Handles are used to create custom legends for the plot."""
     fig, ax = plt.subplots()
     # plot the data
     sns.scatterplot(
--- a/python_scripts/ensemble_gradient_boosting.py
+++ b/python_scripts/ensemble_gradient_boosting.py
@@ -29,16 +29,15 @@
 import pandas as pd
 import numpy as np
 
-rng = np.random.RandomState(0)  # Create a random number generator
-
 
 def generate_data(n_samples=50):
     """Generate synthetic dataset. Returns `data_train`, `data_test`,
     `target_train`."""
     x_max, x_min = 1.4, -1.4
     len_x = x_max - x_min
-    x = rng.rand(n_samples) * len_x - len_x / 2
-    noise = rng.randn(n_samples) * 0.3
+    rng = np.random.default_rng(0)  # Create a random number generator
+    x = rng.uniform(size=(n_samples,)) * len_x - len_x / 2
+    noise = rng.normal(size=(n_samples,)) * 0.3
     y = x**3 - 0.5 * x**2 + noise
 
     data_train = pd.DataFrame(x, columns=["Feature"])
--- a/python_scripts/ensemble_gradient_boosting.py
+++ b/python_scripts/ensemble_gradient_boosting.py
@@ -163,10 +163,10 @@ _ = ax.set_title("Prediction of the initial residuals")
 # this sample in `data_train`.
 
 # %%
-sample = data_train.iloc[[-2]]
+sample = data_train.iloc[[-7]]
 x_sample = sample["Feature"].iloc[0]
-target_true = target_train.iloc[-2]
-target_true_residual = residuals.iloc[-2]
+target_true = target_train.iloc[-7]
+target_true_residual = residuals.iloc[-7]
 
 # %% [markdown]
 # Let's plot the original data, the predictions of the initial decision tree and
--- a/python_scripts/ensemble_gradient_boosting.py
+++ b/python_scripts/ensemble_gradient_boosting.py
@@ -34,9 +34,8 @@ def generate_data(n_samples=50):
     """Generate synthetic dataset. Returns `data_train`, `data_test`,
     `target_train`."""
     x_max, x_min = 1.4, -1.4
-    len_x = x_max - x_min
     rng = np.random.default_rng(0)  # Create a random number generator
-    x = rng.uniform(size=(n_samples,)) * len_x - len_x / 2
+    x = rng.uniform(x_min, x_max, size=(n_samples,))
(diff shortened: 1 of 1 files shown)


python_scripts/ensemble_gradient_boosting.py
```diff
@@ -8,26 +8,28 @@
...
+# In this notebook, we present the gradient boosting decision tree algorithm.
 #
-# Gradient-boosting differs from AdaBoost due to the following reason: instead
-# of assigning weights to specific samples, GBDT will fit a decision tree on the
-# residuals error (hence the name "gradient") of the previous tree. Therefore,
-# each new tree in the ensemble predicts the error made by the previous learner
-# instead of predicting the target directly.
+# Even if AdaBoost and GBDT are both boosting algorithms, they are different in
+# nature: the former assigns weights to specific samples, whereas GBDT fits
+# succesive decision trees on the residual errors (hence the name "gradient") of
```
glemaitre, line 15:
```suggestion
# successive decision trees on the residual errors (hence the name "gradient") of
```

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -8,26 +8,28 @@
 # %% [markdown]
 # # Gradient-boosting decision tree (GBDT)
 #
-# In this notebook, we will present the gradient boosting decision tree
-# algorithm and contrast it with AdaBoost.
+# In this notebook, we present the gradient boosting decision tree algorithm.
```
glemaitre, line 11:
```suggestion
# In this notebook, we present the gradient boosting decision tree (GBDT) algorithm.
```

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -8,26 +8,28 @@
 # %% [markdown]
 # # Gradient-boosting decision tree (GBDT)
```
glemaitre, line 9:
```suggestion
# # Gradient-boosting decision tree
```

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -8,26 +8,28 @@
...
+# Later in this notebook we compare the performance of GBDT (boosting) with that
+# of a Random Forest (bagging) for a particular dataset.
 
 # %%
 import pandas as pd
 import numpy as np
 
-# Create a random number generator that will be used to set the randomness
-rng = np.random.RandomState(0)
+rng = np.random.RandomState(0)  # Create a random number generator
```
glemaitre, line 32: Let's move the generator next to the data generation. We should avoid showing a pattern where people would use the generator across different function and estimators.
So the best practice is just to show it next to the data generation. we could even slightly change the code and have:

```python
def generate_data(n_samples=50, seed=0):
    rng = np.random.default_rng(seed)
    x = rng.normal(size=(n_samples,)) * ...
    noise = rng.normal(size=(s_samples,)) * 0.3
```

using `default_rng` should be encourage nowadays.

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -74,45 +76,61 @@ def generate_data(n_samples=50):
...
 # %%
-# plot the data
-sns.scatterplot(
-    x=data_train["Feature"], y=target_train, color="black", alpha=0.5
-)
-# plot the predictions
-line_predictions = plt.plot(data_test["Feature"], target_test_predicted, "--")
+def plot_decision_tree_with_residuals(y_train, y_train_pred, y_test_pred):
+    # Create a plot and get the Axes object
+    fig, ax = plt.subplots()
```
glemaitre, line 91:
```suggestion
    _fig_, ax = plt.subplots()
```

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -74,45 +76,61 @@ def generate_data(n_samples=50):
...
 
 # %%
-# plot the data
-sns.scatterplot(
-    x=data_train["Feature"], y=target_train, color="black", alpha=0.5
-)
-# plot the predictions
-line_predictions = plt.plot(data_test["Feature"], target_test_predicted, "--")
+def plot_decision_tree_with_residuals(y_train, y_train_pred, y_test_pred):
+    # Create a plot and get the Axes object
```
glemaitre, line 90:
```suggestion
```

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -74,45 +76,61 @@ def generate_data(n_samples=50):
...
+
 
 # %%
-# plot the data
-sns.scatterplot(
-    x=data_train["Feature"], y=target_train, color="black", alpha=0.5
-)
-# plot the predictions
-line_predictions = plt.plot(data_test["Feature"], target_test_predicted, "--")
+def plot_decision_tree_with_residuals(y_train, y_train_pred, y_test_pred):
```
glemaitre, line 89: Add a proper docstring to this function for good practice.

glemaitre: It is particularly important to explain what `handles` will be used for.

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -74,45 +76,61 @@ def generate_data(n_samples=50):
...
 
-# plot the residuals
-for value, true, predicted in zip(
-    data_train["Feature"], target_train, target_train_predicted
-):
-    lines_residuals = plt.plot([value, value], [true, predicted], color="red")
 
-plt.legend(
-    [line_predictions[0], lines_residuals[0]], ["Fitted tree", "Residuals"]
+handles, ax = plot_decision_tree_with_residuals(
```
glemaitre, line 112:
```suggestion
# %%
handles, ax = plot_decision_tree_with_residuals(
```

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -126,29 +144,22 @@ def generate_data(n_samples=50):
...
-    lines_residuals = plt.plot([value, value], [true, predicted], color="red")
-
-plt.legend(
-    [line_predictions[0], lines_residuals[0]],
-    ["Fitted tree", "Residuals"],
-    bbox_to_anchor=(1.05, 0.8),
-    loc="upper left",
-)
-_ = plt.title("Prediction of the previous residuals")
+legend_labels = [
```
glemaitre, line 152: can we pass the legend labels and title as parameter as well like this the `plot_` function takes really care about all plotting.

glemaitre: And we don't need to define the story about the `handles`

python_scripts/ensemble_gradient_boosting.py
```diff
@@ -265,7 +259,9 @@ def generate_data(n_samples=50):
 # second tree corrects the first tree's error, while the third tree corrects the
 # second tree's error and so on).
 #
-# We will compare the generalization performance of random-forest and gradient
+# ## First comparison of GBDT vs random forests
```
glemaitre, line 262:
```suggestion
# ## First comparison of GBDT vs. random forests
```

glemaitre: I'm wondering why "First" here. I would drop it since you might refer to another document.

glemaitre (commented):
Here are some comments.

python_scripts/ensemble_gradient_boosting.py
ArturoAmorQ: I thought that making the helper accept legend labels and titles may be a bit too much of customization.

glemaitre: Fair enough.

PR #765, opened 2024-02-23
FIX Make cross-validation figures less inexact

ArturoAmorQ:
Some people has mentioned in RL that the figures on train-validation-test split and nested cross-validation may be misleading, as they can make the user think the testing samples are used for evaluation **in each cv iteration**.

~~Instead of remaking the figures, this PR proposes a more conservative (but still somewhat inexact) solution by changing the legend labels.~~

**Edit:** This PR now does change the figures to better show the re-train process done by default by the grid search.

cc @lesteve and @tuscland


glemaitre:
Could we order the label into: 1. training, 2. validation (reserved for hyperparameters evaluation), 3. testing (reserved for final or ... evaluation).

ArturoAmorQ:
> Could we order the label into: 1. training, 2. validation (reserved for hyperparameters evaluation), 3. testing (reserved for final or ... evaluation).

Validation sets are not reserved, but actually used in the given cv iteration. In such case there is no big advantage on factorizing the legend labels as 1., 2. and 3.

glemaitre (approved):

PR #766, opened 2024-03-01
MAINT Remove head method

ArturoAmorQ:
As suggested by @ogrisel in RL, this PR removes the `head` method from most of the notebooks when displaying the info of a dataframe. This results in a more vertical notebook but also provides more information in a glance, e.g. the shape of said dataframe.


glemaitre (approved):
LGTM

PR #767, opened 2024-03-06
FIX fix several small typos

thawn:
I fixed several small typos - see the individual commits for details.

### DIFF
--- a/notebooks/cross_validation_ex_01.ipynb
+++ b/notebooks/cross_validation_ex_01.ipynb
@@ -52,7 +52,7 @@
     "exercise.\n",
     "\n",
     "Also, this classifier can become more flexible/expressive by using a so-called\n",
-    "kernel that makes the model become non-linear. Again, no requirement regarding\n",
+    "kernel that makes the model become non-linear. Again, no understanding regarding\n",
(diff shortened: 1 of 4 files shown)


notebooks/cross_validation_ex_01.ipynb
```diff
@@ -52,7 +52,7 @@
     "exercise.\n",
     "\n",
     "Also, this classifier can become more flexible/expressive by using a so-called\n",
-    "kernel that makes the model become non-linear. Again, no requirement regarding\n",
+    "kernel that makes the model become non-linear. Again, no understanding regarding\n",
```
ArturoAmorQ, line 55: Hi @thawn thanks for pointing this out! It certainly is a better wording. Can you also make this change in:
- [`cross_validation_sol_01.py`](https://github.com/INRIA/scikit-learn-mooc/blob/main/python_scripts/cross_validation_sol_01.py)
- [`cross_validation_ex_01.py`](https://github.com/INRIA/scikit-learn-mooc/blob/main/python_scripts/cross_validation_ex_01.py)

In fact, the jupyter notebooks are generated from the python files, so you don't really have to worry about making the changes there.

glemaitre (approved):

PR #768, opened 2024-03-12
Update parameter_tuning_nested.py

psteinb:
inserted missing word

### DIFF
--- a/python_scripts/parameter_tuning_nested.py
+++ b/python_scripts/parameter_tuning_nested.py
@@ -215,7 +215,7 @@ print(f"Accuracy on test set: {accuracy:.3f}")
 # ```
 #
 # However, this evaluation only provides us a single point estimate of the
-# generalization performance. As you recall at the beginning of this notebook, it is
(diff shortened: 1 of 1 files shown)


python_scripts/parameter_tuning_nested.py
```diff
@@ -215,7 +215,7 @@
 # ```
 #
 # However, this evaluation only provides us a single point estimate of the
-# generalization performance. As recall at the beginning of this notebook, it is
+# generalization performance. As you recall at the beginning of this notebook, it is
```
ArturoAmorQ, line 218: Hi @psteinb thanks for the PR! Maybe this wording is a bit better (it may need formatting the line length though):
```suggestion
# generalization performance. As you recall from the beginning of this notebook, it is
```

psteinb: Totally, thanks for lending a second pair of eyes.

ArturoAmorQ (approved):
LGTM! Thanks again @psteinb, merging :rocket:

PR #776, opened 2024-04-23
ENH Improve wording in group-aware cross-validation notebook

ArturoAmorQ:
This PR improves the motivation for group-aware cross-validation techniques, revisits the general wording and changes the number of bins in the plots to better reflect the distributions.

Note: A recurrent question in live sessions is why not using the writers' ids as a feature. This is also addressed here.

### DIFF
--- a/python_scripts/cross_validation_grouping.py
+++ b/python_scripts/cross_validation_grouping.py
@@ -19,6 +19,15 @@ data, target = digits.data, digits.target
 # %% [markdown]
 # We create a model consisting of a logistic regression classifier with a
 # preprocessor to scale the data.
+#
+# ```{note}
+# Here we use a `MinMaxScaler` as we know that each pixel's gray-scale is
+# strictly bounded between 0 (white) and 16 (black). This makes `MinMaxScaler`
+# more suited in this case than `StandardScaler`, as some pixels consistently
+# have low variance (pixels at the borders might almost always be zero if most
+# digits are centered in the image). Then, using `StandardScaler` can result in
+# a very high scaled value due to division by a small number.
+# ```
 
 # %%
 from sklearn.preprocessing import MinMaxScaler
--- a/python_scripts/cross_validation_grouping.py
+++ b/python_scripts/cross_validation_grouping.py
@@ -222,3 +222,7 @@ _ = plt.title("Distribution of the test scores")
 # assessing a model’s ability to generalize to new groups. Without this
 # consideration, the results may appear overly optimistic compared to the actual
 # performance.
+#
+# The interested reader can learn about other group-aware cross-validation
+# techniques in the [scikit-learn user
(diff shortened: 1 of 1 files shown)


python_scripts/cross_validation_grouping.py
```diff
@@ -18,8 +17,8 @@
 data, target = digits.data, digits.target
 
 # %% [markdown]
-# We will recreate the same model used in the previous notebook: a logistic
-# regression classifier with a preprocessor to scale the data.
+# We create a model consisting of a logistic regression classifier with a
+# preprocessor to scale the data.
```
glemaitre, line 21: We might want to explain why not using `StandardScaler`: some pixel have really low variance leading to undesired scale. Instead, we use use case knowledge where a pixel is encoded between 0 and max values and thus `MinMaxScaler` is a better scaling in this case.

python_scripts/cross_validation_grouping.py
```diff
@@ -72,29 +73,29 @@
...
 plt.xlim([0.8, 1.0])
 plt.xlabel("Accuracy score")
 plt.legend(bbox_to_anchor=(1.05, 0.8), loc="upper left")
 _ = plt.title("Distribution of the test scores")
 
 # %% [markdown]
-# The cross-validation testing error that uses the shuffling has less variance
-# than the one that does not impose any shuffling. It means that some specific
-# fold leads to a low score in this case.
+# Shuffling the data results in higher cross-validated test accuracy with less
```
glemaitre, line 88:
```suggestion
# Shuffling the data results in a higher cross-validated test accuracy with less
```

python_scripts/cross_validation_grouping.py
```diff
@@ -72,29 +73,29 @@
...
 # %%
 print(test_score_no_shuffling)
 
 # %% [markdown]
-# Thus, there is an underlying structure in the data that shuffling will break
-# and get better results. To get a better understanding, we should read the
-# documentation shipped with the dataset.
+# Thus, shuffling the data breaks the underlying structure and thus makes the
+# classification task easier to our model. To get a better understanding, we can
+# read the dataset description:
```
glemaitre, line 98:
```suggestion
# read the dataset description in more details:
```

python_scripts/cross_validation_grouping.py
```diff
@@ -176,8 +177,9 @@
 _ = plt.title("Underlying writer groups existing in the target")
 
 # %% [markdown]
-# Once we group the digits by writer, we can use cross-validation to take this
-# information into account: the class containing `Group` should be used.
+# Once we group the digits by writer, we can incorporate this information into
+# the cross-validation process by using group-aware variations of the strategies
+# we have explored in this course, for example, the class `GroupKFold`.
```
glemaitre, line 182:
```suggestion
# we have explored in this course, for example, the `GroupKFold` strategy.
```

glemaitre (commented):
It looks good apart of those little details.

glemaitre (approved):
LGTM

PR #778, opened 2024-05-13
ENH Rework bagging notebook

ArturoAmorQ:
As mentioned in https://github.com/INRIA/scikit-learn-mooc/pull/763#discussion_r1581108888, we should avoid showing a pattern where people would use the random number generator across different function and estimators. This PR fixes the issue and takes the opportunity to rework the general wording.


python_scripts/ensemble_bagging.py
```diff
@@ -127,6 +124,7 @@ def bootstrap_sample(data, target):
     data_bootstrap, target_bootstrap = bootstrap_sample(
         data_train,
         target_train,
+        seed=bootstrap_idx,
```
glemaitre, line 127: I might add a small comment to mention that since we pass different seed, we ensure that the bootstrap samples are different but reproducible.

glemaitre (approved):
LGTM. Thanks @ArturoAmorQ

python_scripts/ensemble_bagging.py
ArturoAmorQ, line 127:
```suggestion
        seed=bootstrap_idx,  # ensure bootstrap samples are different but reproducible
```

PR #780, opened 2024-06-03
FIX Several small fixes

ArturoAmorQ:
1. At the end of the [First look at the dataset notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/01_tabular_data_exploration.html), under the heading "we made important observations" we mention that linear models can only capture linear interactions but there is no prior reference to linear models.

2. The word "complex" at the end of the [Effect of the sample size in cross-validation notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_learning_curve.html) may be misleading. Here we don’t mean complexity as in the depth of a decision tree or the degree of a polynomial (which indeed would lead to overfitting), but rather that a more "expressive" model should be used.

3. In the [Solution for Exercise M3.02](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_sol_03.html) the wording refers to the "accuracy of the model" but the model is a `KNeighborsRegressor`.

4. The `value_counts` function in the [Classification metrics notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/metrics_classification.html) is raising a warning.

This PR proposes some fixes to address those issues.

### DIFF
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -366,6 +366,6 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 #   problem, but may require special treatment as we will cover in future
 #   notebooks;
 # * decision trees create prediction rules by comparing each feature to a
-# threshold value, resulting in decision boundaries that are always parallel to
-# the axes. In 2D, this means the boundaries are vertical or horizontal line
-# segments at the feature threshold values.
+#   threshold value, resulting in decision boundaries that are always parallel
+#   to the axes. In 2D, this means the boundaries are vertical or horizontal
+#   line segments at the feature threshold values.
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -360,9 +360,10 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 # We made important observations (which will be discussed later in more detail):
 #
 # * if your target variable is imbalanced (e.g., you have more samples from one
-#   target category than another), you may need special techniques for training
-#   and evaluating your machine learning model;
-# * having redundant (or highly correlated) columns can be a problem for some
-#   machine learning algorithms;
+#   target category than another), you may need to be careful when interpreting
+#   the values of performance metrics;
+# * columns can be redundant (or highly correlated), which is not necessarily a
+#   problem, but may require special treatment as we will cover in future
+#   notebooks;
 # * decision trees create rules at constant values of the features, such as
 #   vertical and horizontal lines in a 2-dimensional feature space.
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -365,5 +365,7 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 # * columns can be redundant (or highly correlated), which is not necessarily a
 #   problem, but may require special treatment as we will cover in future
 #   notebooks;
-# * decision trees create rules at constant values of the features, such as
-#   vertical and horizontal lines in a 2-dimensional feature space.
+# * decision trees create prediction rules by comparing each feature to a
+# threshold value, resulting in decision boundaries that are always parallel to
+# the axes. In 2D, this means the boundaries are vertical or horizontal line
+# segments at the feature threshold values.
--- a/python_scripts/01_tabular_data_exploration.py
+++ b/python_scripts/01_tabular_data_exploration.py
@@ -364,5 +364,5 @@ _ = plt.annotate("???", (45, 60), fontsize=35)
 #   and evaluating your machine learning model;
 # * having redundant (or highly correlated) columns can be a problem for some
 #   machine learning algorithms;
-# * contrary to decision tree, linear models can only capture linear
-#   interactions, so be aware of non-linear relationships in your data.
+# * decision trees create rules at constant values of the features, such as
+#   vertical and horizontal lines in a 2-dimensional feature space.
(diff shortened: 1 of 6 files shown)


python_scripts/01_tabular_data_exploration.py
```diff
@@ -364,5 +364,5 @@
 #   and evaluating your machine learning model;
 # * having redundant (or highly correlated) columns can be a problem for some
 #   machine learning algorithms;
-# * contrary to decision tree, linear models can only capture linear
-#   interactions, so be aware of non-linear relationships in your data.
+# * decision trees create rules at constant values of the features, such as
+#   vertical and horizontal lines in a 2-dimensional feature space.
```
ogrisel, line 368:
```suggestion
# * decision trees define prediction rules by comparing feature values to thresholds,
#   one feature after another. The resulting decision boundary of a decision tree
#   is therefore composed of vertical and horizontal line segments in a 2-dimensional
#   feature space.
```

python_scripts/01_tabular_data_exploration.py
```diff
@@ -364,5 +364,5 @@
 #   and evaluating your machine learning model;
```
ogrisel, line 364: => "you may need to be careful when interpreting the values of performance metrics: for instance, the accuracy of a baseline prediction rule that constantly predicts the majority class can be much larger than 50%."

python_scripts/01_tabular_data_exploration.py
```diff
@@ -364,5 +364,5 @@
...
 # * having redundant (or highly correlated) columns can be a problem for some
 #   machine learning algorithms;
```
ogrisel, line 366:
```suggestion
# * columns can be redundant (or highly correlated). This is not necessary a problem
#   for machine learning models but it can make it challenging to assess the
#   relative importance of the input variables in their ability to make accurate
#   predictions.    
```

ogrisel (commented):
Thanks, here are some further improvements:

python_scripts/01_tabular_data_exploration.py
ArturoAmorQ: Thanks for the review @ogrisel. I addressed your other comments, but this wording feels a bit too complex for an introductory notebook.

python_scripts/cross_validation_learning_curve.py
```diff
@@ -104,7 +104,7 @@
 #
 # If we achieve a plateau and adding new samples in the training set does not
 # reduce the testing error, we might have reached the Bayes error rate using the
-# available model. Using a more complex model might be the only possibility to
+# available model. Using a more expressive model might be the only possibility to
 # reduce the testing error further.
```
ArturoAmorQ, line 108: @ogrisel If a model reached Bayes error then isn’t that error is irreducible meaning if we even select a more expressive model that won’t improve our test error?

ArturoAmorQ: Wording fixed in [bcc45f3](https://github.com/INRIA/scikit-learn-mooc/pull/780/commits/bcc45f3adb6df161780ab042d9b222e2d7259066).

python_scripts/01_tabular_data_exploration.py
ogrisel: > Thanks for the review @ogrisel. I addressed your other comments, but this wording feels a bit too complex for an introductory notebook.

Personally, I find the previous wording more confusing: "create rules at constant values of the features" does not mean anything. What does a "rule at a value" mean?

Maybe the following is both simpler and actually meaningful:

```suggestion
# * decision trees define prediction rules that successively compare each feature to
#   a threshold value. The resulting decision boundary of a decision tree is
#   therefore composed of vertical and horizontal line segments in a 2-dimensional
#   feature space.
```

python_scripts/cross_validation_learning_curve.py
```diff
@@ -102,10 +102,11 @@
...
 #
-# If we achieve a plateau and adding new samples in the training set does not
-# reduce the testing error, we might have reached the Bayes error rate using the
-# available model. Using a more complex model might be the only possibility to
-# reduce the testing error further.
+# If the testing error plateaus despite adding more training samples, it's
+# possible that the model has achieved its optimal performance. In this case,
+# using a more expressive model might help reduce the error further. Otherwise,
+# the error may have reached the Bayes error rate, the theoretical minimum error
+# due to inherent uncertainty in the data.
```
ogrisel, line 109: It might be helpful to explain why this is expected:

```suggestion
# due to inherent uncertainty not resolved by the available data. This minimum error is
# non-zero whenever some of the variation of the target variable `y` depends on
# external factors not fully observed in the features available in `X`, which is
# almost always the case in practice.
```

ogrisel (approved):
Some more feedback. LGTM otherwise.

ogrisel:
@ArturoAmorQ I finally answered your feedback in https://github.com/INRIA/scikit-learn-mooc/pull/780#discussion_r1853981184.

python_scripts/01_tabular_data_exploration.py
```diff
@@ -360,9 +360,12 @@
...
-#   interactions, so be aware of non-linear relationships in your data.
+#   target category than another), you may need to be careful when interpreting
+#   the values of performance metrics;
+# * columns can be redundant (or highly correlated), which is not necessarily a
+#   problem, but may require special treatment as we will cover in future
+#   notebooks;
+# * decision trees create prediction rules by comparing each feature to a
+# threshold value, resulting in decision boundaries that are always parallel to
+# the axes. In 2D, this means the boundaries are vertical or horizontal line
+# segments at the feature threshold values.
```
ArturoAmorQ, line 371:
```suggestion
# * decision trees create prediction rules by comparing each feature to a
#   threshold value, resulting in decision boundaries that are always parallel
#   to the axes. In 2D, this means the boundaries are vertical or horizontal
#   line segments at the feature threshold values.
```

PR #783, opened 2024-11-26
remove dublicate word

psteinb:
Just discovered while teaching this notebook.


ArturoAmorQ (approved):
Good catch, thanks @psteinb!

PR #784, opened 2024-11-26
better formulation to wrap-up

psteinb:
The last sentence of the `cross_validation_time.ipynb` notebook felt a bit convoluted. I tried my best to improve the formulation.

### DIFF
--- a/python_scripts/cross_validation_time.py
+++ b/python_scripts/cross_validation_time.py
@@ -218,7 +218,8 @@ test_score = cross_val_score(
 print(f"The mean R2 is: {test_score.mean():.2f} ± {test_score.std():.2f}")
 
 # %% [markdown]
-# In conclusion, it is really important to not use an out of the shelves
+# In conclusion, it is really important not to carelessly use a
 # cross-validation strategy which do not respect some assumptions such as having
-# i.i.d data. It might lead to absurd results. These absurd results can make you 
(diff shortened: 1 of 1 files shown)


python_scripts/cross_validation_time.py
```diff
@@ -220,5 +220,5 @@
 # %% [markdown]
 # In conclusion, it is really important to not use an out of the shelves
```
ArturoAmorQ, line 221: I don't think "out of the shelves" is grammatically correct nor a common term for the students to know. Maybe we can rephrase the whole paragraph.

ArturoAmorQ (commented):
Thanks for the PR @psteinb! Here are a couple of comments. Also, the linter fail in the CI is because it's configured to 80 characters max.

Being said that, reworking this whole notebook has been in my list of to do's for a long time. In particular, we want to use a more realistic dataset, give interpretation to resulting R2 and MSE scores, and mention the actual good practices for modeling, e.g. aligning the test size of `TimeSeriesSplit` with the forecasting task. In case you want to contribute, feel free to submit a more in depth PR.

python_scripts/cross_validation_time.py
ArturoAmorQ, line 221:
```suggestion
# In conclusion, it is really important not to carelessly use a
```

PR #785, opened 2024-12-09
Use balanced accuracy with adult census dataset

ArturoAmorQ:
Inspired by [this question in the forum](https://mooc-forums.inria.fr/moocsl/t/class-imbalance-handling/15868), this PR uses balanced accuracy instead of accuracy to score a logistic regression pipeline with an interaction feature engineering step.

It feels more natural to me to use balanced accuracy as scoring metric, as it has already been used for the adult census dataset in previous notebooks.

I also took the opportunity to tweak the general wording.

### DIFF
--- a/python_scripts/linear_models_sol_03.py
+++ b/python_scripts/linear_models_sol_03.py
@@ -24,10 +24,13 @@ data
 # %% [markdown]
 # We confirm that all the selected features are numerical.
 #
-# Compute the generalization performance in terms of accuracy of a linear model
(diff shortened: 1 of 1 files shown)


glemaitre:
I would not recommend to use the balanced accuracy because it is an arbitrary metric. I would not advise it anymore in scikit-learn as well.

Here, I don't think that the balanced ratio is that catastrophic: we could therefore compare the accuracy of a `DummyClassifier` predicting the majority class and mention that we need any predictive model to perform above this accuracy level to ensure that the model capture any signal.

PR #788, opened 2024-12-12
MAINT Improve build of exercises to avoid duplicated 'Write your code here' cells

ArturoAmorQ:
Sometimes we split the solution code into more than a single cell. That creates consecutive "Write your code here cells".

This PR improves the build of exercises from their solution to avoid such duplicates.


ogrisel (approved):
LGTM. Can you re-run this script as part of this PR to see the diff and check that it works as expected before merging?

Issue #651, opened 2022-07-12
How to cite this course/repo?


psteinb:
In order to pay tribute to this fantastic content, I'd love to cite this repo or this course. It would be wonderful, if you could share in the `README.md` how to cite this.
Thanks in advance!

lesteve:
Thanks for your kind words, it would be nice to add "how to cite this course" to the README indeed.

Having said that, I don't really know how other course repos manage this kind of thing. Suggestions more than welcome!

I am aware of this https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content but not sure we want to have a DOI for particular releases.

psteinb:
I would encourage you to have DOIs ... for every release if possible.

lesteve:
A few links I bumped into that may serve as data points:
- https://the-turing-way.netlify.app/welcome#citing-the-turing-way
- https://the-environmental-ds-book.netlify.app/preamble/attribution.html

For now we don't really do releases since the content is regularly evolving but we could if we think this is worth it (e.g. for each MOOC session) ...

lesteve:
I followed the instructions in https://docs.github.com/en/repositories/archiving-a-github-repository/referencing-and-citing-content. My understanding is that we need to create a github release for Zenodo to create automatically a DOI.

Remaining steps:
- do a github relase for the MOOC 3rd version
- add a small "how to cite us" section in `jupyter-book/index.md` similarly as in https://inria.github.io/scikit-learn-mooc/

psteinb:
Great news! Thanks for making the effort. This will make it easier for us to cite this material.

PR #802, opened 2025-02-24
MNT Add CITATION.cff

lesteve:
Follow-up from https://github.com/INRIA/scikit-learn-mooc/issues/651. I tested CITATION.cff in a test repo and this seems to do the job.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,5 +1,5 @@
 cff-version: 1.2.0
-message: "If you use this software, please cite it as below."
+message: "If you use this content, please cite it as below."
 authors:
   - name: "The scikit-learn MOOC developers"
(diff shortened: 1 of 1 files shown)


CITATION.cff
```diff
@@ -0,0 +1,8 @@
+cff-version: 1.2.0
+message: "If you use this software, please cite it as below."
```
ArturoAmorQ, line 2: Maybe a slightly more appropriate wording.
```suggestion
message: "If you use this content, please cite it as below."
```

ArturoAmorQ (approved):
Thanks @lesteve, merging!

ArturoAmorQ:
It seems that the change in the `message` didn't work as intended:

image

PR #790, opened 2024-12-12
MAINT Update scikit-learn to v 1.6

ArturoAmorQ:
Fixes #789.

Issue #789, opened 2024-12-12
Bump scikit-learn version to 1.6 on JupyterHub


ArturoAmorQ:
We currently use scikit-learn v1.3, but #772 requires v1.4. Updating scikit-learn to its latest version (v1.6) will allow us to give a better maintenance to the mooc.

On FUN, updating the Jupyter image of the students seems to suffice.

lesteve:
Not sure about the implications on the FUN MOOC if any, but I'll let you deal with this :wink:

lesteve:
Thinking about it more, I guess for now none of our notebooks actually need scikit-learn 1.6 but oh well ...

ArturoAmorQ:
> Not sure about the implications on the FUN MOOC if any, but I'll let you deal with this 😉

Pinging @brospars for the Jupyter image update :)

> Thinking about it more, I guess for now none of our notebooks actually need scikit-learn 1.6 but oh well

Not necessarily 1.6, but https://github.com/INRIA/scikit-learn-mooc/pull/772 needs 1.4, (until now, we used to run on 1.3)

lesteve:
> Pinging @brospars for the Jupyter image update :)

I seem to remember in FUN-MOOC the environment is created when the user starts its first notebook. I don't remember if there was a way for the user to update its environment but @brospars is indeed the one that may know :wink:.

> until now, we used to run on 1.3

This does not matter too much but just to be 100% explicit. 1.3 is our minimum version requirement `scikit-learn>=1.3` meant that we were using the latest release, for example in the CI or if someone new come to the repo and follows our installation instructions.

brospars:
The notebook image is this one : https://hub.docker.com/r/brospars/singleuser-scikit-learn-mooc (source : https://github.com/inrialearninglab/singleuser-scikit-learn-mooc)

I disabled auto updates last year to prevent potential errors due to breaking changes. 

If I run the pipeline to update dependencies. This will run these lines :

``` Dockerfile
FROM jupyter/scipy-notebook:notebook-6.5.4
# [...]
# Install mooc requirements
RUN wget https://raw.githubusercontent.com/INRIA/scikit-learn-mooc/master/requirements.txt
RUN python3 -m pip install --no-cache -r requirements.txt
```

And every user will have the updated version upon the next startup of their environment.

Do I proceed ?

lesteve:
[...]

PR #810, opened 2025-03-21
MNT Fix several FutureWarnings

ArturoAmorQ:
Follows from https://github.com/INRIA/scikit-learn-mooc/pull/790.

Bumping the version scikit-learn to v1.6 introduced some FutureWarnings. This PR fixes them, as well as a `FutureWarning` related to resampling time indexed data.

Note: As mentioned in https://github.com/INRIA/scikit-learn-mooc/issues/789#issuecomment-2575302965, auto-updates are disabled on FUN, where still run on scikit-learn v1.3. Therefore building the notebooks after this PR should be done in combination with an update of the Jupyter image in FUN.

### DIFF
--- a/python_scripts/parameter_tuning_grid_search.py
+++ b/python_scripts/parameter_tuning_grid_search.py
@@ -89,6 +89,8 @@ from sklearn.compose import ColumnTransformer
 preprocessor = ColumnTransformer(
     [("cat_preprocessor", categorical_preprocessor, categorical_columns)],
     remainder="passthrough",
+    force_int_remainder_cols=False,  #  avoid warning in scikit-learn v1.6
+    # we make explicit that remainder columns are not forced to be integers
 )
 
 # %% [markdown]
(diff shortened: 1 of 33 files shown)


python_scripts/parameter_tuning_grid_search.py
```diff
@@ -89,6 +89,9 @@
 preprocessor = ColumnTransformer(
     [("cat_preprocessor", categorical_preprocessor, categorical_columns)],
     remainder="passthrough",
+    # Silence a deprecation warning in scikit-learn v1.6 related to how
+    # the `transformers_` attribute stores column identifiers internally.
```
ogrisel, line 93:
```suggestion
    # Silence a deprecation warning in scikit-learn v1.6 related to how this
    # class stores an internal attribute that we do not use in this notebook.
```

python_scripts/parameter_tuning_grid_search.py
```diff
@@ -89,6 +89,9 @@
 preprocessor = ColumnTransformer(
     [("cat_preprocessor", categorical_preprocessor, categorical_columns)],
     remainder="passthrough",
+    # Silence a deprecation warning in scikit-learn v1.6 related to how the
+    # ColumnTransformer stores an attribute that we do not use in this notebook
```
glemaitre, line 93: Just to confirm, I assume it is middle ground between @ogrisel and @ArturoAmorQ

glemaitre (approved):

Issue #792, opened 2024-12-23
Revisit explanation on category encoding strategies


fritshermans:
I think that the statement on which category encoding strategy to choose could use some revision:

_"In general OneHotEncoder is the encoding strategy used when the downstream models are linear models while OrdinalEncoder is often a good strategy with tree-based models."_

There is no problem in using the OneHotEncoder with a tree based model as long as there are not too many unique categorical values. Moreover, I don't think it's good to teach that tree based models may be presented with ordinal encoded categorical values where the order doesn't make sense because the tree will be able to sort things out (pun intended) when they are deep enough. Imagine a modelling exercise where a country's climate is an important predictor and the country name is a categorical feature; 'Australia' (warm), 'Austria' (cold), 'Barbados' (warm), 'Belgium' (cold) etc. The tree needs to be very deep only because of poor categorical value encoding.

I propose to add the TargetEncoder with some explanation. This is a better suited encoder when there are many unique categories and there is no logical order.

ArturoAmorQ:
Hi @fritshermans thanks for pointing this out!

Concerning your comment on adding `TargetEncoder`, we were planning to add a brief discussion on the native support of categorical values in hist-boosted trees. Adding yet another way of encoding can be a bit too much information for this relatively introductory course. What we sometimes do is just add a short message at the end of a lesson, e.g. in the last 2 paragraphs of the [Regression notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/metrics_regression.html), where we point interested readers to the documentation.

Similarly we can add a small comment in the [Importance of decision tree hyperparameters on generalization notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_hyperparameters.html) about ordinal encoding when there are too many unique categorical values.

Don't hesitate to open a PR!

fritshermans:
Sure, I will do that! Thanks for your time to check my remarks.

fritshermans:
I think this issue could be closed by [PR 811](https://github.com/INRIA/scikit-learn-mooc/pull/811)

ArturoAmorQ:
Thanks!

PR #811, opened 2025-03-30
Add tip to read about TargetEncoder in the documentation

fritshermans:
As proposed by @ArturoAmorQ in issue [#792, ](https://github.com/INRIA/scikit-learn-mooc/issues/792), I added a note to read about the TargetEncoder. Please let me know if this is what you had in mind.


ArturoAmorQ:
In https://github.com/INRIA/scikit-learn-mooc/issues/792#issuecomment-2619296610 I suggested adding a small comment in the [Importance of decision tree hyperparameters on generalization notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/trees_hyperparameters.html) about ordinal encoding when there are too many unique categorical values.

On a second thought, I think your current contribution and the comment on the tree `max_depth` are more suited for the [Wrap-up analysis in the Exercise M1.05](https://inria.github.io/scikit-learn-mooc/python_scripts/03_categorical_pipeline_sol_02.html#id1).

Notice there we already have a small comment "Trees can deal with `OrdinalEncoder` fine as long as they are deep enough." that can be easily overlooked at, so maybe we can:
- take the table off from the `Important` banner;
- improve the entry for `Tree-based model` vs `Non-meaningful order` to something similar to "`OrdinalEncoder` with a reasonable depth";
- move the paragraph in this PR at the very end of this notebook;
- maybe remove the `tags=["solution"]` so that this whole wrap-up appears in the exercise and not only in it's solution.

fritshermans:
Thanks for your remarks @ArturoAmorQ!

I applied your suggestions. The only thing that I would propose to change is to add some explanation on 'reasonable depth' before the table. Moreover, why would a `OneHotEncoder` result in 10x more features than an `OrdinalEncoder`? That depends on the number of categories right?

Please let me know what you think!

python_scripts/03_categorical_pipeline_sol_02.py
```diff
@@ -170,7 +170,7 @@
     f"with a fitting time of {elapsed_time:.3f}"
 )
 
-# %% [markdown] tags=["solution"]
+# %% [markdown]
```
ArturoAmorQ, line 173: Sorry for the second thoughts, but I now realize it's weird to show the interpretation of the results in the exercise. We can rather split a bit further below, to display the conclusions starting from the "Which encoder should I use" section.

```suggestion
# %% [markdown] tags=["solution"]
```

python_scripts/03_categorical_pipeline_sol_02.py
```diff
@@ -190,14 +190,15 @@
 # perfectly fine for `HistGradientBoostingClassifier` and yields fast training
 # times.
 
-# %% [markdown] tags=["solution"]
-# ```{important}
 # Which encoder should I use?
```
ArturoAmorQ, line 193:
```suggestion
# %% [markdown]
# ## Which encoder should I use?
```

ArturoAmorQ (commented):
Just a nitpick on the current changes.

Otherwise I think you are right with respect to the comment on `OneHotEncoder` resulting in 10x more features than an `OrdinalEncoder`. It does depend on the cardinality.

And also you are right that we should explain that "reasonable depth" means. This can be done by completing the phrase "Trees can deal with OrdinalEncoder fine as long as they are deep enough [to sort things out without overfitting...]" inside the `Important` banner.

fritshermans:
Done! Let me know what you think of it 😊

ArturoAmorQ (approved):
I took the liberty to sync the notebooks in your PR, hope you don't mind :)

Thanks again @fritshermans for your time and enthusiasm. Merging!

fritshermans:
thanks @ArturoAmorQ ! What should I have done for syncing the notebooks? They didn't show up as 'changed' by git on my local computer I think

ArturoAmorQ:
You can use either `make quizzes` or just `make` to make all of the notebooks, or you can make one notebook at the time as documented in [Contributing -Updating the notebooks ](https://github.com/INRIA/scikit-learn-mooc/blob/main/CONTRIBUTING.md#updating-the-notebooks).

It's just that we prefer to keep reviews focused on the python files, and then make the notebooks on our side at the end. The downside of such workflow is having too many PRs like https://github.com/INRIA/scikit-learn-mooc/pull/824 and https://github.com/INRIA/scikit-learn-mooc/pull/825 :sweat_smile:

Issue #809, opened 2025-03-21
Use `data_train` in validation curve


ArturoAmorQ:
As reported in [this forum post](https://mooc-forums.inria.fr/moocsl/t/use-of-data-or-data-train/15880), in the instructions for the validation curve in the [Exercise M6.03](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_sol_03.html) it says 
>For both the gradient-boosting and random forest models, create a validation curve using the training set to assess the impact of the number of trees on the performance of each model.

But then in the solution we don't use the `data_train` but all data for the validation curve.

Related to #603, we can take the opportunity to also emphasize at this stage that `n_estimators` should not be tuned for random forests, and that the validation curve in the notebook was only computed for educational purposes.

fritshermans:
@ArturoAmorQ, as the HistGradientBoostingClassifier/Regressor by default applies only auto-stopping when there are at least 10k samples, it would make sense to do explain this behavior?

ArturoAmorQ:
You are right, that would make sense. Maybe we can additionally point students towards this piece of documentation: [Effect of number of trees and early stopping](https://scikit-learn.org/stable/auto_examples/ensemble/plot_hgbt_regression.html#effect-of-number-of-trees-and-early-stopping).

fritshermans:
Apologies @ArturoAmorQ, I realize now that a `GradientBoostingClassifier` is used, not the histogram based version of it. Let me add the validation curve for GBT as that is what the exercise prescibes.

fritshermans:
I think this issue could be closed by [PR 827](https://github.com/INRIA/scikit-learn-mooc/pull/827)

ArturoAmorQ:
You are right. Thanks for pointing this out.

PR #812, opened 2025-03-31
[WIP] Use data_train in validation curve

fritshermans:
This PR aims to solve issue https://github.com/INRIA/scikit-learn-mooc/issues/809:

- [ ] use data_train and target_train for ValidationCurveDisplay
- [ ] emphasize that n_estimators should not be tuned

### DIFF
--- a/python_scripts/ensemble_sol_03.py
+++ b/python_scripts/ensemble_sol_03.py
@@ -69,8 +69,8 @@ from sklearn.model_selection import ValidationCurveDisplay
 param_range = np.array([1, 2, 5, 10, 20, 50, 100])
 disp = ValidationCurveDisplay.from_estimator(
     forest,
-    data,
-    target,
+    data_train,
+    target_train,
     param_name="n_estimators",
     param_range=param_range,
     scoring="neg_mean_absolute_error",


ArturoAmorQ:
Hi @fritshermans thanks for the PR! About the emphasis that `n_estimators` should not be tuned, currently we have a message in the notebook [Hyperparameter tuning](https://inria.github.io/scikit-learn-mooc/python_scripts/ensemble_hyperparameters.html), which comes two notebooks after this exercise:
```
# ```{caution}
# Tuning the `n_estimators` for random forests generally result in a waste of
# computer power. We just need to ensure that it is large enough so that doubling
# its value does not lead to a significant improvement of the validation error.
# ```
```
We can add a similar caution message in the lines of:
> As we will see more in detail in the following section, tuning the `n_estimators` for random forests generally results in a waste of computer power. The validation curve in the present exercise was only computed for educational purposes.

(As a side note, it seems that there is a typo in "generally result(s)" in the mentioned message.)

ogrisel (approved):
Let's fix the two problems separately. I merged `main` and ran `make exercises notebooks` to check the rendering of this PR on the CI.

I think we can already merge this fix and discuss the other problem in a follow-up PR if needed.

PR #819, opened 2025-04-02
ENH Add note about link between `DummyRegressor(strategy="mean")` and R²

### LINKED ISSUE #806: Add interpretation of vanishing R² score in terms of `DummyRegressor`
In the [Comparing model performance with a simple baseline notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_baseline.html) we introduce the `DummyRegressor` as baseline using MAE.

I think it's an opportunity to mention that, from the [mathematical definition of the R² score](https://scikit-learn.org/stable/modules/model_evaluation.html#r2-score-the-coefficient-of-determination), a model that always predicts the mean results in a zero training R² score. Therefore, the strategy="mean" of the [`sklearn.dummy.DummyRegressor`](https://scikit-learn.org/stable/modules/generated/sklearn.dummy.DummyRegressor.html) can be considered an implicit baseline for such score.

Having this explanation would help interpreting the scores in the [Non i.i.d data notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_time.html), then I consider this issue to be part of #786.

augustebaum:
Closes #806


python_scripts/cross_validation_baseline.py
```diff
@@ -75,6 +75,47 @@
...
+# an R² score higher than 0 then it performs better
+# than a `DummyRegressor` with `strategy="mean"`; similarly, if the
+# R² score is lower than 0 then your model is worse than the dummy
+# regressor.
+
+# %% [markdown]
+# ```{admonition} Mathematical explanation
+# :class: dropdown
+# Recall that the R² score is defined as
+# $$
```
glemaitre, line 105: It does not render as expected. It looks like it only consider a single `$`

ogrisel: I don't know if it can support multiline equations. Here is the rendering from the CI preview:

image

https://pull-request-819--scikit-learn-mooc.netlify.app/python_scripts/cross_validation_baseline

ogrisel: The documentation says it should work:

https://jupyterbook.org/en/stable/content/math.html#math-blocks

python_scripts/cross_validation_baseline.py
```diff
@@ -75,6 +75,47 @@
...
+# R^2 = 1 - \frac{\sum_i (y_i - \hat{y}_i)^2}{\sum_i (y_i - \bar{y})^2}
+# $$
+#
+# But our model always predicts the mean, i.e. for all $i$,
+# $\hat{y}_i = \bar{y}$, so
+# $$
```
ogrisel, line 111: Let's try this maybe?

```suggestion
# $\hat{y}_i = \bar{y}$, so:
#
# $$
```

python_scripts/cross_validation_baseline.py
ogrisel, line 105: and that:

```suggestion
# Recall that the R² score is defined as:
#
# $$
```

ogrisel: Maybe it's a bad interaction with the use of the admonition?

python_scripts/cross_validation_baseline.py
```diff
@@ -75,6 +75,49 @@
...
+# Recall that the R² score is defined as:
+#
+# $$
+# R^2 = 1 - \frac{\sum_i (y_i - \hat{y}_i)^2}{\sum_i (y_i - \bar{y})^2}
+# $$
+#
+# But our model always predicts the mean, i.e. for all $i$,
+# $\hat{y}_i = \bar{y}$, so:
+#
+# $$
```
ogrisel, line 113:
```suggestion
# $
```

python_scripts/cross_validation_baseline.py
```diff
@@ -75,6 +75,49 @@
...
+# R^2 = 1 - \frac{\sum_i (y_i - \hat{y}_i)^2}{\sum_i (y_i - \bar{y})^2}
+#     = 1 - \frac{\sum_i (y_i - \bar{y}  )^2}{\sum_i (y_i - \bar{y})^2}
+#     = 1 - 1
+#     = 0.
+# $$
```
ogrisel, line 118:
```suggestion
# $
```

python_scripts/cross_validation_baseline.py
```diff
@@ -75,6 +75,49 @@
...
+# regressor.
+
+# %% [markdown]
+# ```{admonition} Mathematical explanation
+# :class: dropdown
+# Recall that the R² score is defined as:
+#
+# $$
+# R^2 = 1 - \frac{\sum_i (y_i - \hat{y}_i)^2}{\sum_i (y_i - \bar{y})^2}
+# $$
```
ogrisel, line 108:
```suggestion
# $
```

python_scripts/cross_validation_baseline.py
```diff
@@ -75,6 +75,49 @@
...
+# than a `DummyRegressor` with `strategy="mean"`; similarly, if the
+# R² score is lower than 0 then your model is worse than the dummy
+# regressor.
+
+# %% [markdown]
+# ```{admonition} Mathematical explanation
+# :class: dropdown
+# Recall that the R² score is defined as:
+#
+# $$
```
ogrisel, line 106:
```suggestion
# $
```

ogrisel (commented):
Let's try to see out the rendering looks like with single dollar signs. Maybe it's good enough.

python_scripts/cross_validation_baseline.py
augustebaum: I went to look at the CI preview (link above) and it looks like this:

2025-04-03T10_41_33_screenshot

augustebaum: Now it looks like there's a weird vertical space before the first equation

ArturoAmorQ:
I find weird to have a progression where we speak about MAE, then R2, then back to MAE. I would move your contribution to either:
- the very beginning of the notebook and then say we change to MAE to have a natural way to compare units or
- at the very end, as a side comment like "By the way, we could have used R2 isntead, as it induces an implicit benchmark, but what you earn in interpretability of the sign, you lose in interpretability of the units".

ogrisel:
+1 for moving at the end. We could also do the histogram of r2 score values for the model vs dummy.

augustebaum:
Done

python_scripts/cross_validation_baseline.py
ogrisel: So removing the double dollar appears to be fine.

Interestingly, on firefox, here is what I get:

image

I think this is good enough either way.

ogrisel (approved):
LGTM this way (with or without doing the histogram for the r2 score).

ArturoAmorQ:
Oups, I was about to tweak a couple of details :sweat_smile:. I will open another PR.

Issue #786, opened 2024-12-10
Rework Non i.i.d. data notebook


ArturoAmorQ:
The plot showcasing the use of a `ShuffleSplit` strategy  [Non i.i.d. data notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_time.html) changed in a recent version of the pandas plotting utility:

**In previous versions:**
![shuffle_1](https://github.com/user-attachments/assets/0488131a-2406-43d9-ad55-75fafaa9dff8)

**Now:**
![shuffle_2](https://github.com/user-attachments/assets/527e9bdb-c298-4176-8622-dd3821bf8c01)

We could use the opportunity to rework this whole notebook, as mentioned in https://github.com/INRIA/scikit-learn-mooc/pull/784#pullrequestreview-2484855966:

- Not use groups with `TimeSeriesSplit` (it's currently rising a `UserWarning`)
- Use a more realistic dataset (optional)
- Give interpretation to resulting R2 and MSE scores:
  - results are above or below chance level?
  - they are over optimistic when not evaluated properly
  - predictions are not realistic, when using the current dataset, a simple `DecisionTreeRegressor` can foresee a sudden drop in quotes
- Mention the actual good practices for modeling, e.g. aligning the test size of `TimeSeriesSplit` with the forecasting task
- In general give more focus to the use of `TimeSeriesSplit`

glemaitre:
I wanted to report this issue because I just saw the plot. A hot fix for the plot is to make sure that the data point are ordered by increasing date.

PR #826, opened 2025-04-03
Iterate on the interpretation of R2 score

ArturoAmorQ:
Follow-up from #819. See also #786.

### DIFF
--- a/python_scripts/cross_validation_baseline.py
+++ b/python_scripts/cross_validation_baseline.py
@@ -114,7 +114,7 @@ _ = plt.title("Cross-validation testing errors")
 # case for dataset with extreme outliers.
 
 # %% [markdown]
-# Finally, let us see what happens if we measure the test score using R²
+# Finally, let us see what happens if we measure the test score using $R^2$
 # instead of the mean absolute error:
 
 # %%
@@ -127,13 +127,13 @@ r2_train_score_dummy_regressor = pd.Series(
 r2_train_score_dummy_regressor.describe()
 
 # %% [markdown]
-# The R² score is always 0. It can be shown that this is always the case,
+# The $R^2$ score is always 0. It can be shown that this is always the case,
 # because of its mathematical definition. If you are interested in the proof,
 # unfold the dropdown below.
 #
 # ```{admonition} Mathematical explanation
 # :class: dropdown
-# Recall that the R² score is defined as:
+# Recall that the $R^2$ score is defined as:
 #
 # $
 # R^2 = 1 - \frac{\sum_i (y_i - \hat{y}_i)^2}{\sum_i (y_i - \bar{y})^2}
@@ -150,11 +150,12 @@ r2_train_score_dummy_regressor.describe()
 # $
 # ```
 #
-# This helps put your model's R² score in perspective: if your model has an R²
-# score higher than 0 then it performs better than a `DummyRegressor` with
-# `strategy="mean"`; similarly, if the R² score is lower than 0 then your model
-# is worse than the dummy regressor. For the test score, we observe something
-# similar, but with an additional effect coming from the dataset variations:
+# This helps put your model's $R^2$ score in perspective: if your model has an
+# $R^2$ score higher than 0 then it performs better than a `DummyRegressor` with
+# `strategy="mean"`; similarly, if the $R^2$ score is lower than 0 then your
+# model is worse than the dummy regressor. For the test score, we observe
+# something similar, but with an additional effect coming from the dataset
+# variations:
 
 # %%
 r2_test_score_dummy_regressor = pd.Series(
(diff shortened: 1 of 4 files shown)


python_scripts/cross_validation_baseline.py
```diff
@@ -114,29 +114,23 @@
...
     dummy, data, target, cv=cv, scoring="r2", return_train_score=True, n_jobs=2
 )
 r2_train_score_dummy_regressor = pd.Series(
-    result_dummy["train_score"], name="Dummy regressor"
+    result_dummy["train_score"], name="Dummy regressor train score"
 )
 r2_train_score_dummy_regressor.describe()
 
 # %% [markdown]
 # The R² score is always 0. It can be shown that this is always the case,
```
auguste-probabl, line 130:
```suggestion
# The R² score is close to 0. It can be shown that this is always the case,
```
I remember we tried it together

python_scripts/cross_validation_baseline.py
```diff
@@ -155,3 +149,21 @@
...
+r2_test_score_dummy_regressor = pd.Series(
+    result_dummy["test_score"], name="Dummy regressor test score"
+)
+r2_test_score_dummy_regressor.describe()
+
+# %% [markdown]
+# In conclusion, R² establishes an implicit baseline, but while it provides
+# interpretability regarding whether the model performs better or worse than the
+# dummy regressor that always predicts the mean of the target, it sacrifices
+# interpretability in terms of the units.
```
ogrisel, line 169: I think that R2 is quite interpretable, even if its scaled is independent of the scale of the target variable.

So let me suggest:


```suggestion
# In conclusion, R² is a normalized performance score independent
# of the physical unit of the target: 0.0 is the performance of a regressor
# that constantly predicts the mean observed value of the target variable and
# 1.0 is the performance of a perfect model that predicts exactly the observed
# target variable for each given input observation. Note that it is only
# possible to reach 1.0 if the target variable is a deterministic function of
# the available input features. In most practical settings, some variations of
# the target variable are determined by external factors independent of the
# input variables. So for a given feature set, the R² score of the optimal
# predictive model is strictly below 1.0, not because of a problem with the
# model itself, but because the available observed feature values are not
# informative enough to deterministically determine the value of the target
# variable.
#
# All in all, R² can be intuitively interpreted as the fraction of the 
# variations of the target variable that the model can successfully predict
# using the available input features.
```

python_scripts/cross_validation_baseline.py
augustebaum, line 130:
```suggestion
# The R² score is close to 0. It can be shown that this is always the case,
```
I think we tried this together

python_scripts/cross_validation_time.py
```diff
@@ -176,9 +176,9 @@
 print(f"The R2 on this single split is: {test_score:.2f}")
 
 # %% [markdown]
-# In this case, we see that our model is not magical anymore. Indeed, it
-# performs worse than just predicting the mean of the target. We can visually
-# check what we are predicting.
+# In this case, we see that our model is not magical anymore. Remember that a
+# negative $R^2$ means that the regressor performs worse than always predicting
```
ogrisel, line 180: nitpick: sometimes we use `$R^2$` while other times we use `R²`. We should probably try to stay consistent, at least within a single md file.

python_scripts/cross_validation_time.py
```diff
@@ -176,9 +176,9 @@
...
+# the mean of the target. We can visually check what we are predicting.
```
ogrisel, line 181:
```suggestion
# the mean of the target. We can visually check what we are predicting as follows:
```

ogrisel (approved):
Thanks for the follow-up. Here are some further suggestions.

Other than that, +1 for merge after a sync.

python_scripts/cross_validation_baseline.py
```diff
@@ -155,3 +150,33 @@
...
+# physical unit of the target variable, unlike MAE. A $R^2$ score of 0.0 is the
+# performance of a model that always predicts the mean observed value of the
+# target, while 1.0 corresponds to a model that predicts exactly the observed
+# target variable for each given input observation. Notice that it is only
+# possible to reach 1.0 if the target variable is a deterministic function of
+# the available input features. In practice, external factors often introduce
+# variability in the target that cannot be explained by the available features.
+# Therefore, the $R^2$ score of an optimal model is typically less than 1.0, not
+# due to a fail in the model, but because the input features alone are
+# insufficient to deterministically predict the target.
```
ogrisel, line 178:
```suggestion
# due to a limitation of the machine learning algorithm itself, but because
# the chosen input features are fundamentally not informative enough to be
# able to deterministically predict the target.
```

python_scripts/cross_validation_baseline.py
```diff
@@ -155,3 +150,33 @@
 #     = 0.
 # $
 # ```
+#
+# This helps put your model's $R^2$ score in perspective: if your model has an
+# $R^2$ score higher than 0 then it performs better than a `DummyRegressor` with
+# `strategy="mean"`; similarly, if the $R^2$ score is lower than 0 then your
+# model is worse than the dummy regressor. For the test score, we observe
+# something similar, but with an additional effect coming from the dataset
+# variations:
```
ogrisel, line 159:
```suggestion
# variations: the mean target value measured on the testing set is slightly
# different from the mean target value measured on the training set.
```

ogrisel (approved):
Final feedback ;)

PR #827, opened 2025-04-10
Add validation curve for n_estimators plus advise on (not) tuning

fritshermans:
As discussed in [issue 809](https://github.com/INRIA/scikit-learn-mooc/issues/809), I add the advise not to tune `n_estimators`. Moreover, I add the validation curve for this hyperparameter as that is what is asked for in the exercise. Please feel free to adjust the wording as you like.


ArturoAmorQ (approved):
Perfect, thanks again! Merging!

Issue #829, opened 2025-04-18
Use MAE isntead of R2 in Exercise M3.02


ArturoAmorQ:
[Exercise M3.02](https://inria.github.io/scikit-learn-mooc/python_scripts/parameter_tuning_sol_03.html) uses the default R2 score to evaluate a model for the first time, but we only explain how to interpret R2 in [this notebook from M7](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_baseline.html). At this stage we've already introduced MAE in [this notebook form M2](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_train_test.html) (also using California Housing), so it is more natural to use MAE at this stage.

Changing R2 to MAE would require:
- updating the instructions to ask setting `scoring="neg_mean_absolute_error"`;
- remind the student to introduce a minus sign to the "mean test score" column;
- update the analysis by the end of the solution to correctly use MAE.

fritshermans:
I can pick this one up if you don't mind

PR #830, opened 2025-04-19
replace R^2 with MAE in exercise M3.02

fritshermans:
This PR aims to close [#829](https://github.com/INRIA/scikit-learn-mooc/issues/829). Let me know your feedback!


fritshermans:
I get a linter error that I don't understand. @ArturoAmorQ, could you please have a look what goes wrong here?

python_scripts/parameter_tuning_sol_03.py
```diff
@@ -69,8 +74,8 @@
 from sklearn.model_selection import RandomizedSearchCV
 
 param_distributions = {
-    "kneighborsregressor__n_neighbors": np.logspace(0, 3, num=10).astype(
-        np.int32
+    "kneighborsregressor__n_neighbors": (
+        np.logspace(0, 3, num=10).astype(np.int32)
```
ArturoAmorQ, line 78: This should fix the CI:
```suggestion
    "kneighborsregressor__n_neighbors": np.logspace(0, 3, num=10).astype(
        np.int32
```

ArturoAmorQ (commented):
Thanks for addressing this issue @fritshermans!
Here's just a quick pass to your PR.

python_scripts/parameter_tuning_sol_03.py
```diff
@@ -50,6 +50,11 @@
 # - the parameter `with_std` of the `StandardScaler` with possible values `True`
 #   or `False`.
 #
+# Use mean absolute error as model performance metric. Remember that a score
+# function should always return higher values for better performance.
+# Therefore, we should choose the negative version as the score metric:
+# `neg_mean_absolute_error`.
```
ArturoAmorQ, line 56: I would even be very explicit by changing L43-L44 to something similar to
> Use `RandomizedSearchCV` with `n_iter=20` and `scoring="neg_mean_absolute_error"` to tune the following hyperparameters of the `model`:

and later (as you did) give a more detailed explanation in the lines of:
```suggestion
# The `scoring` function is expected to return higher values for better models,
# since grid/random search objects **maximize** it. Because of that, error
# metrics like `mean_absolute_error` must be negated (using the `neg_` prefix)
# to work correctly (remember lower errors represent better models).
```

ArturoAmorQ (commented):
Here's a more in depth review. Feel free to rephrase my wording if you can simplify it.

python_scripts/parameter_tuning_sol_03.py
```diff
@@ -106,6 +113,12 @@
 import pandas as pd
 
 cv_results = pd.DataFrame(model_random_search.cv_results_)
+# %% [markdown]
```
ArturoAmorQ, line 116: >The exercise only asks for printing the best hyperparameter combination. The solution is more extensive by showing the parallel coordinates plot etc so that's why I didn't add it to the exercise. I suggest to either add to the exercise to show the test score for the optimal hyperparameter combination or to leave it as is, but then there is no need to add the instruction of multiplying by -1 to the exercise as well. Please let me know your thoughts.

Yes, you are right, I apologize for having read too fast. Then let's just add the `solution` tag (and a blank line for spacing).
```suggestion

# %% [markdown] tags=["solution"]
```

fritshermans:
I think we're set and done 😊

fritshermans:
It's a pleasure!

Issue #452, opened 2021-09-06
Inconsistent use of `ColumnTransformer` and `make_column_transformer`


ArturoAmorQ:
Most of the notebooks use `ColumnTransformer`

glossary.md
predictive_modeling_module_take_away.md:
wrap_up_quiz 1
03_categorical_pipeline_column_transformer.py
03_categorical_pipeline_ex_02.py
03_categorical_pipeline_sol_02.py
parameter_tuning_ex_02.py and parameter_tuning_sol_02.py
parameter_tuning_ex_03.py and parameter_tuning_sol_03.py
parameter_tuning_grid_search.py
parameter_tuning_nested.py
parameter_tuning_randomized_search.py

On the other hand, `make_column_transformer` is only used in

ensemble_random_forest.py
wrap_up_quiz 1
wrap_up_quiz 4
wrap_up_quiz 5

I recommend keeping either one or the other to avoid possible confusion.
Keeping `make_column_transformer` means more editing in the present contents but will reduce variance naming (and therefore calling) the transformers. What do you think?

glemaitre:
> I recommend keeping either one or the other to avoid possible confusion.

I would use both but make sure that we present both. We should have the same issue with `Pipeline` and `make_pipeline`. I see that we do a slightly better job stating that `make_pipeline` is creating a `Pipeline`. But I think it would be worth showing how to create a `Pipeline` directly.

The reason to use both is that there are time where you want to use one or another, specifically.

ArturoAmorQ:
Solved in #831.

PR #831, opened 2025-04-23
MAINT Changed the use of ColumnTransformer to make_column_transformer

SebastienMelo:
Per issue #452, made the use of make_column_transformer more uniform in the MOOC. Since the name of the preprocessors isn't used to modify their parameters, it seems simpler to teach the students to create a ColumnTransformer with the function.


ogrisel (approved):
LGTM, I will sync with main and regenerate the notebooks and then merge if all green.

PR #832, opened 2025-04-30
MTN Sample grouping digits

SebastienMelo:
As per [this question](https://mooc-forums.inria.fr/moocsl/moocsl/t/confusion-with-the-documentation/15906), made the explaination in the notebook clearer that `load_digits`only takes the test part of the handwritten digits dataset.


python_scripts/cross_validation_grouping.py
```diff
@@ -111,12 +111,13 @@
 
 # %% [markdown]
 # If we read carefully, 13 writers wrote the digits of our dataset, accounting
```
ArturoAmorQ, line 113:
```suggestion
# If we read carefully, `load_digits` loads a copy of the **test set** of the
```

python_scripts/cross_validation_grouping.py
```diff
@@ -111,12 +111,13 @@
...
-# training or the testing sets. Mixing the data will break this structure, and
-# therefore digits written by the same writer will be available in both the
-# training and testing sets.
+# for a total amount of 1797 samples. Indeed, `load_digits` loads only the test
+# data of original dataset. Thus, a writer wrote several times the same numbers.
+# Let's suppose that the writer samples are grouped. Subsequently, not shuffling
+# the data will keep all writer samples together either in the training or the
+# testing sets. Mixing the data will break this structure, and therefore digits
+# written by the same writer will be available in both the training and testing
+# sets.
```
ArturoAmorQ, line 120:
```suggestion
# UCI ML hand-written digits dataset, which consists of 1797 images wrote by
# **13 different writers**. Thus, each writer wrote several times the same
# numbers. Let's suppose the dataset is ordered by writer. Subsequently,
# not shuffling the data will keep all writer samples together either in the
# training or the testing sets. Mixing the data will break this structure, and
# therefore digits written by the same writer will be available in both the
# training and testing sets.
```

ArturoAmorQ (commented):
Thanks for pointing this out and for the PR @SebastienMelo! I just rephrased a little bit. WDYT?

Sorry my suggestion is divided into two, but github doesn't allow for a single suggestion on both modified and non-modified lines.

python_scripts/cross_validation_grouping.py
SebastienMelo: Yes, seems better to me!

python_scripts/cross_validation_grouping.py
```diff
@@ -110,9 +110,10 @@
 print(digits.DESCR)
 
 # %% [markdown]
-# If we read carefully, 13 writers wrote the digits of our dataset, accounting
-# for a total amount of 1797 samples. Thus, a writer wrote several times the
-# same numbers. Let's suppose that the writer samples are grouped. Subsequently,
+# If we read carefully, `load_digits` loads a copy of the **test set** of the
+# UCI ML hand-written digits dataset, which consists of 1797 images wrote by
```
ArturoAmorQ, line 114:
```suggestion
# UCI ML hand-written digits dataset, which consists of 1797 images by
```

ArturoAmorQ (approved):
Perfect,then it LGTM. Thanks for the contribution @SebastienMelo! Merging!

PR #834, opened 2025-05-14
MTN Switched the use of cross validation and train_test_split in non iid data notebook

SebastienMelo:
As per [this question](https://mooc-forums.inria.fr/moocsl/moocsl/t/a-specific-question/15908), this PR switches the order in which the cross validation and the train test split happen to make it more coherent with the rest of the notebook and the modules.


python_scripts/cross_validation_time.py
```diff
@@ -103,7 +92,7 @@
 from sklearn.model_selection import cross_val_score
 
 test_score = cross_val_score(
```
ArturoAmorQ, line 94:
```suggestion
test_score = cross_val_score(regressor, data, target, cv=cv, n_jobs=2)
```

python_scripts/cross_validation_time.py
```diff
@@ -103,7 +92,7 @@
...
-    regressor, data_train, target_train, cv=cv, n_jobs=2
+    regressor, data, target, cv=cv, n_jobs=2
 )
```
ArturoAmorQ, line 96:
```suggestion
```

ArturoAmorQ (approved):
Just a tweak but otherwise LGTM. Thanks @SebastienMelo!

PR #849, opened 2025-07-09
MTN Bias variance quizz

SebastienMelo:
Synchronized the bias variance quizz with gitlab.


ArturoAmorQ (approved):
Thanks @SebastienMelo!

PR #853, opened 2025-08-07
MTN fixed the parallel plots

SebastienMelo:
Per [this question](https://mooc-forums.inria.fr/moocsl/t/parallel-coordinates-not-working/15922), added the fix `renderer="notebook"` where it was missing.


ArturoAmorQ (approved):

PR #860, opened 2025-09-10
FIX HistGradientBoosting fitting time too long

SebastienMelo:
As per [this question](https://mooc-forums.inria.fr/moocsl/t/fitting-time-of-histgradientboostingregressor/15930) related to [this issue](https://github.com/scikit-learn/scikit-learn/issues/30662), commented the n_jobs parameters so that the results make sense.


ArturoAmorQ:
Thanks again @SebastienMelo! Merging!

PR #871, opened 2026-01-14
MTN Exercise M4.01 target correction

SebastienMelo:
As per [this question](https://mooc-forums.inria.fr/moocsl/t/dataframes-vs-series-or-why-youre-getting-nans/15955), fixes the target definition to `penguins[[target_name]]` to prevent code complications in the solution. The solution is therefore simpler.


ArturoAmorQ (approved):
Thanks @SebastienMelo!

PR #876, opened 2026-05-04
Add module on dimensionality reduction

ArturoAmorQ:
Adds module on PCA. Adds the wiki_news dataset, notebooks and exercises.
Probably a good idea to merge this after the clustering module in #836.

Note: This is still WIP. We are missing at least one quiz and the wrap-up quiz.


python_scripts/dimred_intuitions.py
```diff
@@ -0,0 +1,343 @@
...
+# %%
+fig, ax = plt.subplots()
+bars = ax.barh(
+    range(1, len(pca.explained_variance_) + 1),
+    pca.explained_variance_.round(decimals=1),
+)
+ax.bar_label(bars)
+ax.set_xlim([0, 14])
+ax.set_yticks([1, 2], labels=["PC1", "PC2"])
+ax.set_xlabel("eigenvalues")
```
fritshermans: My suggestion would be to use explained variance in the plot as it the term eigenvalues is not used (yet):


```suggestion
ax.set_xlabel("Explained variance")
```

python_scripts/dimred_preprocessing.py
```diff
@@ -0,0 +1,586 @@
...
+# Looking at the heatmap after scaling, you might wonder if the loadings being
+# evenly distributed is just a consequence of `StandardScaler` giving every
+# feature unit variance. Does that mean all features contribute more or less
+# equally to every component? Not quite. Each feature axis has unit variance,
+# yes. But PCA looks for directions in the **joint space** of all features. The
+# variance along a diagonal direction depends on how much the features correlate
+# with each other. Two strongly correlated features will jointly define a
+# direction with much higher variance than either one alone, and the explained
+# variance ratios across components will still be very unequal. Scaling removes
+# the unit bias; it does not make all directions equally important.
```
fritshermans: I would say that without scaling a magnitude bias is introduced more than a unit bias:

```suggestion
# the magnitude bias; it does not make all directions equally important.
```

python_scripts/dimred_ex_01.py
```diff
@@ -0,0 +1,202 @@
...
+#       format_name: percent
+#       format_version: '1.3'
+#       jupytext_version: 1.17.0
+#   kernelspec:
+#     display_name: Python 3
+#     name: python3
+# ---
+
+# %% [markdown]
+# # Solution for Exercise M8.01
```
fritshermans:
```suggestion
# # Exercise M8.01
```

python_scripts/dimred_text.py
```diff
@@ -0,0 +1,424 @@
...
+print(f"Kaiser criterion: {n_kaiser} components")
+
+# %% [markdown]
+# Both heuristics suggest keeping far more components than is practical. The 90%
+# variance threshold already requires on the order of thousands of components.
+# The Kaiser criterion, which sets the threshold at 1/6678 explained
+# variance, keeps even more: all 900 components we computed pass it, meaning the
+# true cutoff lies beyond what we measured.
+#
+# For text data, a common practice is to fix the number of components to be 100
```
fritshermans, line 224: Out of curiousity; where do these values come from? Why not between 100 and 300?

ArturoAmorQ: I also hesitated about that. My logic was as follows:
- When doing hyperparameter search, it is common practice to explore values that are roughly equally spaced on a log scale (1, 3, 5, 10, 30, 50, 100, 300, 500, ...). Under that convention, 100 and 300 are the values that make the most sense when working with text data.
- For LSA specifically, 100 components is a well-established default. 300 is the next log-spaced candidate, so it serves as a sensible upper bound to explore.

I don't want to imply that trying 150 or 200 would be wrong, but I wouldn't expect a meaningful difference between 100 and 150. WDYT of a wording in the lines of:
```
# For text data, a common practice is to fix the number of components to be 100
# or 300. Other values are valid, but a finer grid is unlikely to yield
# meaningfully different results and would add unnecessary computational cost.
```

fritshermans: aaaah, that makes total sense! I like how you phrase the explanation; compact and clear!

python_scripts/dimred_intuitions.py
```diff
@@ -0,0 +1,343 @@
...
+
+penguins = pd.read_csv("../datasets/penguins_classification.csv")
+penguins = penguins[penguins["Species"] == "Chinstrap"]
+penguins = penguins.drop(columns="Species")
+penguins.plot.scatter(x="Culmen Length (mm)", y="Culmen Depth (mm)")
+
+# %% [markdown]
+# ## Finding the principal component
+#
+# PCA is intended to find new features (called principal components, or "PC")
```
jeromedockes, line 34: I think it would help to clarify the vocabulary and describe each of the objects we are talking about, giving for each of them its shape and its different names (eg V^T is the `components_` attribute, called principal axes, of shape n_components x n_input_features, the output of `transform()` is `XV`, ...) otherwise it can be easy to get confused about what is what. when talking about directions it can be helpful to say in which space.

I realize there is a tension between this and keeping the explanation short and at a high-level for now, but I think it would help to fix a bit ideas and would not hesitate to repeat the sizes of the things we are talking about

python_scripts/dimred_intuitions.py
```diff
@@ -0,0 +1,343 @@
...
+# that capture enough of the structure in our data.
+#
+# The first PC is the direction along which our data varies the most. Because
+# our features are correlated, there is a clear dominant pattern in the data
+# which PCA can identify. The second PC would align with the direction with
+# second to most variance, and so on.
+#
+# In our case we begin by extracting both components to understand the full
+# picture. **This is not dimensional reduction yet**, these components are
+# linear combinations of the original features, in other words, just a rotation
```
jeromedockes: would it help to use terms like 'coordinate system' to explain the different spaces and how they relate?

python_scripts/dimred_intuitions.py
```diff
@@ -0,0 +1,343 @@
...
+    alpha=0.6,
+    ax=ax,
+)
+ax.plot(x1_range, x2_pred, "b-", linewidth=2, label="Regression line")
+
+ax.plot(pc_line[:, 0], pc_line[:, 1], "r-", linewidth=2, label="First PC")
+
+ax.set_title("Regression line vs First PC")
+ax.legend()
+plt.show()
```
jeromedockes: i think it would help to actually draw the lines representing the errors. ie draw the vertical line from the point to the approximation in one color and say that is the residual, and draw the line perpendicular to the pc direction in another color and say that is the error minimized by pca

python_scripts/dimred_preprocessing.py
```diff
@@ -0,0 +1,592 @@
...
+# directions that reflect the combined structure of all three features, not just
+# the one with the largest raw variance. The overall shape looks different
+# because the two pipelines found genuinely different directions in the original
+# space, not just a rescaled version of the same projection.
+#
+# To see exactly which original features each PC captures, we can read the
+# `components_` attribute. It is a an array of shape `(n_components,
+# n_features)`. Each entry of the array is the cosine of the angle between the
+# axis of an original feature and one PC direction. That is, a value close to 0
+# means the PC is nearly perpendicular to the original feature, whereas a value
```
fcharras:
```suggestion
# means the PC is nearly perpendicular to the original feature axis, whereas a value
```

python_scripts/dimred_components.py
```diff
@@ -0,0 +1,367 @@
...
+# and various phenolic compounds. The goal in the original task is to identify
+# the cultivar from the chemistry alone.
+#
+# We use it here purely to practice selecting the number of PCA components on
+# real data where the features have very different scales and units.
+
+# %%
+from skrub import TableReport
+
+TableReport(X)
```
SebastienMelo:
```suggestion
TableReport(X, verbose=0)
```

python_scripts/dimred_intuitions.py
```diff
@@ -0,0 +1,387 @@
...
+    (penguins.values @ pca.components_.T).var(axis=0, ddof=1)
+)  # ddof scales by n_samples-1
+
+# %% [markdown]
+# Since the variance is the square of the standard deviation, we have to take
+# the square root to recover the original scale in millimeters:
+
+# %%
+pca_std = np.sqrt(pca.explained_variance_)
+print(f"Std along the first PC : {pca_std[0]:.3f} mm")
```
SebastienMelo, line 149: I find it weird to give the unit here since PCA can mix units. Maybe add a comment like "in the general case the std along a PC has no unit because they can mix several heterogeneous units".

python_scripts/dimred_preprocessing.py
```diff
@@ -0,0 +1,592 @@
...
+    sq_loadings = pipe[-1].components_ ** 2
+    im = plot_sq_loadings(ax, pipe[-1], features_4d)
+    ax.set_title(title)
+fig.colorbar(im, ax=axes, shrink=0.8)
+plt.show()
+
+# %% [markdown]
+# With `StandardScaler`, PC1 loads on `GrLivArea` and `OverallQual` the most
+# (positively correlated house-size and quality features), while `LotArea` and
+# `YearBuilt` contribute less to PC1, and more to PC2. The outlier lots do
```
SebastienMelo: On the plot `YearBuilt` is more "present" in the first PC. Rephrase by changing to something like "contribute less to PC1 **relatively to** `GrLivArea` ..."

python_scripts/dimred_ex_01.py
```diff
@@ -0,0 +1,202 @@
...
+fig.update_layout(
+    title={
+        "text": "Trade-off between fit time and mean test score",
+        "y": 0.95,
+        "x": 0.5,
+        "xanchor": "center",
+        "yanchor": "top",
+    }
+)
+fig.show(renderer="notebook")
```
SebastienMelo: Define as a function that the student will copy paste to prevent the apparition of errors in the notbook

python_scripts/dimred_components.py
```diff
@@ -0,0 +1,367 @@
...
+# counts under two pipelines: one using the Kaiser recommendation
+# (`n_components=3`) and one using the 90% threshold (`n_components=8`).
+#
+# As in the clustering notebook, we plot the silhouette score against number of
+# clusters and look for an elbow. Notice that silhouette is not comparable
+# *across* different numbers of components, since the score depends on the space
+# in which distances are computed, even if the score is normalized. We are not
+# comparing the two curves against each other, instead we are checking whether
+# each curve, read on its own, points to the same number of clusters.
+#
```
SebastienMelo, line 296: Explain why we are not computing it in the original space maybe.

python_scripts/dimred_components.py
```diff
@@ -0,0 +1,367 @@
...
+ax.set_xlabel("Number of components")
+ax.set_ylabel("Cumulative explained variance")
+ax.set_title("Cumulative variance is not identical across subsamples")
+_ = ax.legend()
+
+# %% [markdown]
+# Each faint line is one subsample. The spread is most visible near the
+# threshold lines, which means a fixed number of components does not guarantee
+# exactly 90% or 95% variance retained. We can quantify this directly.
+
```
SebastienMelo, line 177: Rephrase to say that it is different from 90%, because we select an integer of dimensions. The phrasing is a bit odd with respect to the plot because there seems to be very little spread near the threshold lines.

python_scripts/dimred_components.py
```diff
@@ -0,0 +1,367 @@
...
+    ax.set_xlabel("Number of clusters (n_clusters)")
+    ax.set_ylabel("Silhouette score")
+    ax.set_title(label)
+
+# %% [markdown]
+# The two panels tell a different story. With the Kaiser components, the peak at
+# 3 clusters is clear and consistent across all subsamples. With 8 components
+# the curves are much more spread out and there is no clear consensus.
+#
```
SebastienMelo, line 343: Add a comment about the fact that we manage to estimate overfitting with the PCA in an unsupervised manner

SebastienMelo (commented):
Here are the few comments we discussed, thank you @ArturoAmorQ !

python_scripts/dimred_preprocessing.py
```diff
@@ -0,0 +1,592 @@
...
```
SebastienMelo:
```suggestion
# Using `StandardScaler`, PC1 is dominated by `OverallQual` and `GrLivArea`, with
# `YearBuilt` contributing moderately; while `LotArea` plays almost no role in
# PC1 but dominates PC2. The outlier lots do
```

SebastienMelo (approved):
Everything seems good to me!

ArturoAmorQ:
As this PR is becoming too large, I suggest we merge it as it is (with @SebastienMelo's approve) and we can iterate later. In particular, we still need to add the wrap-up quiz.

PR #878, opened 2026-06-17
MTN Added a note on using stratification for cross validation

### LINKED ISSUE #875: Revisit the Stratification notebook
The [Stratification notebook](https://inria.github.io/scikit-learn-mooc/python_scripts/cross_validation_stratification.html) concludes:

> Stratifying is more effective than just shuffling when it comes to making sure that the distributions of classes in all the folds are representative of the entire dataset. As training and testing folds have similar class distributions, stratifying leads to a more realistic measure of the model’s ability to generalize. This is specially important when the performance metrics depend on the proportion of the positive class, as we will see in a future notebook.

In https://github.com/scikit-learn/scikit-learn/pull/30576 a note was added mentioning the impact of stratification on the variability of the performance metric:

> Stratification makes cross-validation folds more homogeneous, and as a result hides some of the variability inherent to fitting models with a limited number of observations.
>
> As a result, stratification can artificially shrink the spread of the metric measured across cross-validation iterations: the inter-fold variability does no longer reflect the uncertainty in the performance of classifiers in the presence of rare classes.

The message in the notebook should make this point clearer.

For more context, see [this notebook](https://gist.github.com/ogrisel/af21bdc55a2c02671a48c68631ee7294) by @ogrisel.

SebastienMelo:
Fixes #875. This PR proposes adding a note to the stratification notebook.

### DIFF
--- a/python_scripts/cross_validation_stratification.py
+++ b/python_scripts/cross_validation_stratification.py
@@ -265,12 +265,10 @@ _ = plt.title("Test set class counts\n(with stratifying)")
 #
 # However, as noted in the [scikit-learn user
 # guide](https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation-iterators-with-stratification-based-on-class-labels),
-# stratification makes folds more homogeneous, which can artificially shrink the
-# spread of the performance metric across folds in a classification case where
-# classes are heavily imbalanced. The reduced inter-fold variability no longer
-# reflects the true uncertainty in model performance — it is a consequence of
-# the folds being deliberately similar, not a sign that the model is more
-# stable.
+# stratification makes the folds more homogeneous. In the presence of severe
+# class imbalance, this can artificially reduce the variability of performance
+# metrics across folds, causing the observed variability to underestimate the
+# true uncertainty in model performance.
 #
 # The interested reader can learn about other stratified cross-validation
 # techniques in the [scikit-learn user
--- a/python_scripts/cross_validation_stratification.py
+++ b/python_scripts/cross_validation_stratification.py
@@ -255,13 +255,11 @@ _ = plt.title("Test set class counts\n(with stratifying)")
 # train set and the test set. The difference is due to the small number of
 # samples in the iris dataset.
 #
-# In other words, stratifying is more effective than just shuffling when it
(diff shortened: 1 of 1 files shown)


python_scripts/cross_validation_stratification.py
```diff
@@ -263,6 +263,15 @@
...
 # see in a future notebook.
 #
+# However, as noted in the [scikit-learn user
+# guide](https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation-iterators-with-stratification-based-on-class-labels),
+# stratification makes folds more homogeneous, which can artificially shrink the
+# spread of the performance metric across folds in a classification case where
+# classes are heavily imbalanced. The reduced inter-fold variability no longer
+# reflects the true uncertainty in model performance — it is a consequence of
+# the folds being deliberately similar, not a sign that the model is more
+# stable.
```
ArturoAmorQ: A bit cleaner wording:

```suggestion
# However, as noted in the [scikit-learn user
# guide](https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation-iterators-with-stratification-based-on-class-labels),
# stratification makes the folds more homogeneous. In the presence of severe
# class imbalance, this can artificially reduce the variability of performance
# metrics across folds, causing the observed variability to underestimate the
# true uncertainty in model performance.
```

python_scripts/cross_validation_stratification.py
```diff
@@ -263,6 +263,15 @@
 # performance metrics depend on the proportion of the positive class, as we will
 # see in a future notebook.
```
ArturoAmorQ: I don't know why but the github interface doesn't seem to remove R258 to R264, but the whole paragraph is to be replaced:

```suggestion
# Stratification is especially useful for ensuring that rare classes are
# represented in every cross validation split. In particular, if a class is
# absent from one or more splits, some classification metrics may become
# undefined.
```

ArturoAmorQ (approved):
Thanks @SebastienMelo !

python_scripts/cross_validation_stratification.py
SebastienMelo: Remark taken into account in the next commit

PR #880, opened 2026-07-23
Fix minor formatting issues

StefRe:
Fix typos.

### DIFF
--- a/notebooks/01_tabular_data_exploration.ipynb
+++ b/notebooks/01_tabular_data_exploration.ipynb
@@ -368,8 +368,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "For every entry in `\\\"education\\\"`, there is only one single corresponding\n",
-    "value in `\\\"education-num\\\"`. This shows that `\"education\"` and\n",
+    "For every entry in `\"education\"`, there is only one single corresponding\n",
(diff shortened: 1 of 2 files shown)


ArturoAmorQ:
Hi @StefRe, thanks for contributing! Can you please also modify the corresponding python file?

https://github.com/INRIA/scikit-learn-mooc/blob/0cc70c41c640578c28eb7fd38d8c1c19e09c8aa6/python_scripts/01_tabular_data_exploration.py#L234-L235

StefRe:
Done. I wasn't aware that there were also Python scripts that need to be kept in sync.  

Just out of curiosity: why are both versions maintained in version control, rather than generating the Python scripts automatically from the notebooks? What are the Python scripts even for—they hardly produce any output, do they?

ArturoAmorQ (approved):
> I wasn't aware that there were also Python scripts that need to be kept in sync.

It's ok, no need to know everything to make a valuable contribution.

> Just out of curiosity: why are both versions maintained in version control, rather than generating the Python scripts automatically from the notebooks?

It's the other way around: the jupyter notebooks are generated from the python files using jupytext, because ipynb files tend to get polluted with metadata when edited, and that makes version control a bit less fun of an experience.

But thanks again @StefRe! Merging!

